PR #174, opened 2023-03-14
feat: include pending certificates to source head

atanmarko:
# Description

Get source head position request to TCE now also includes pending certificates. On restart, sequencer will continue producing certificates from the last certificate previously submitted to the TCE (may it be already delivered or still pending).

Fixes TP-436

## Additions and Changes

SourceStreamPosition now has optional position member, because pending certificates do not have position until they are delivered.

## Breaking changes

TCE GRPC API updated and cleaned up. Duplicate TCE SourceStreamPosition removed from tce proto file.

## PR Checklist:


atanmarko:
After discussion on the topic, we decided to keep the existing implementation of `get_source_head` function that returns only delivered certificate. We will introduce another TCE API function that would return pending certificates.

PR #184, opened 2023-03-17
feat: add instrumentation for sequencer

atanmarko:
# Description

Setup `jaeger`/`otel` instrumentation for `sequencer`.
Added  `NewCertificate` telemetry record.
Performed a bit of cleanup in `ceritification` crate.

Fixes # (issue)
TP-333

## Additions and Changes

Now both `sequencer` and `tce` node share same parameters for jaeger agent and jaeger service name.
If they are not provided, only local log tracing will be performed.

### Breaking changes

Changed `topos cli` command line parameters.
Changed environment variable for otlp(jaeger) agent and otlp service name to `TOPOS_OTLP_AGENT` and `TOPOS_OTLP_SERVICE_NAME`.

## PR Checklist:

### DIFF
--- a/crates/topos/src/components/sequencer/commands/run.rs
+++ b/crates/topos/src/components/sequencer/commands/run.rs
@@ -40,12 +40,12 @@ pub struct Run {
 
     /// Socket of the opentelemetry agent endpoint
     /// If not provided open telemetry will not be used
-    #[arg(long, env = "TOPOS_OTLP_AGENT", global = true)]
+    #[arg(long, env = "TOPOS_OTLP_AGENT")]
     pub otlp_agent: Option<String>,
 
     /// Otlp service name
     /// If not provided open telemetry will not be used
-    #[arg(long, env = "TOPOS_OTLP_SERVICE_NAME", global = true)]
+    #[arg(long, env = "TOPOS_OTLP_SERVICE_NAME")]
     pub otlp_service_name: Option<String>,
 }
 
--- a/crates/topos/src/components/tce/commands/run.rs
+++ b/crates/topos/src/components/tce/commands/run.rs
@@ -51,12 +51,12 @@ pub struct Run {
 
     /// Socket of the opentelemetry agent endpoint
     /// If not provided open telemetry will not be used
-    #[arg(long, env = "TOPOS_OTLP_AGENT", global = true)]
+    #[arg(long, env = "TOPOS_OTLP_AGENT")]
     pub otlp_agent: Option<String>,
 
     /// Otlp service name
     /// If not provided open telemetry will not be used
-    #[arg(long, env = "TOPOS_OTLP_SERVICE_NAME", global = true)]
+    #[arg(long, env = "TOPOS_OTLP_SERVICE_NAME")]
     pub otlp_service_name: Option<String>,
 }
 


atanmarko:
CC @sebastiendan environment variables for telemetry have been changed

crates/topos-sequencer-tce-proxy/src/lib.rs
```diff
@@ -491,11 +492,11 @@ impl TceProxyWorker {
...
                         match cmd {
-                            TceProxyCommand::SubmitCertificate(cert) => {
-                                info!(
-                                    "Submitting new certificate to the TCE network: {:?}",
-                                    &cert
-                                );
+                            TceProxyCommand::SubmitCertificate{cert, ctx} => {
+                                let span = info_span!("Sequencer TCE Proxy");
+                                span.set_parent(ctx);
+                                let _span = span.enter();
```
Freyskeyd, line 498: You should not use `enter` in an async method because it will produce incorrect traces.
You should use `entered` method:
```suggestion
                                let _span = span.entered();
```

or even `in_scope`:
```suggestion
                                span.in_scope(|| { .. });
```

crates/topos/src/main.rs
```diff
@@ -29,8 +29,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     LogTracer::init()?;
 
     let args = options::Opt::parse();
-
-    let mut layers = Vec::new();
+    let jaeger_agent = args.jaeger_agent;
```
Freyskeyd, line 32: We can rename every `jaeger` by `otlp` or `opentelemetry` I guess. We will not use `jaeger` anymore

crates/topos/src/main.rs
```diff
@@ -63,19 +65,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         },
     );
 
-    #[cfg(feature = "tce")]
-    if let options::ToposCommand::Tce(TceCommand {
-        subcommands: Some(TceCommands::Run(ref cmd)),
-        ..
-    }) = args.commands
-    {
+    if let (Some(jaeger_agent), Some(jaeger_service_name)) = (jaeger_agent, jaeger_service_name) {
```
Freyskeyd, line 68: Here you removed the filter on the command and put the OTLP argument globally, but I don't want telemetry to be setup for every `topos` commands. Just the one needed, which are sequencer and TCE.
If we stay like this, every command with the environment variables will push traces/logs to our collector.

atanmarko: No, if you do not specify it by default otlp is not initialized.

Freyskeyd: Yes but if I have the environment variables define it will enable it on every command. What I want to say is that it is better to activate the telemetry only for some commands and not on the presence of a variable.

atanmarko: We have cli commands for sequencer and tce. How many commands will actually we have? 
I think better solution is to error on otlp set for some unwanted command (exceptional case) then to duplicate all the code for telemetry setup.

atanmarko: @hadjiszs What do you think?

Freyskeyd: For now, we have those commands that do not need `OTLP`:

- `Tce::status`
- `Tce::push_peer_list`
- `Tce::keys`

In the next PR we will have verifier and a lot more coming in the next iterations.
In fact, the commands that need `OTLP` are `Tce::Run` and `Sequencer::Run` I guess.

To deliver a good user experience with the `topos` cli we should handle everything for the user. If the user is executing a command with some variables defined, we should not fail the command and ask him to do something if they are completely unneeded.

Having the `OTLP` on the root options means that they display on every command too, misleading the user that they can define them when they are not used.

atanmarko: Ok, fair enough. I will set them up per sequencer and tce command.

Freyskeyd (approved):

atanmarko:
@topos-network/tools Just to note `TOPOS_OTLP_SERVICE_NAME` and `TOPOS_OTLP_AGENT` are now used instead of `TCE_JAEGER_SERVICE_NAME` and `TCE_JAEGER_AGENT`

PR #187, opened 2023-03-20
feat: add a new way to put custom labels/keyvalue into the OTLP

Freyskeyd:
# Description

To be able to add custom fields to the `OTLP` logs/traces, the `TOPOS_OTLP_TAGS` variable is defined and will be read to extract those resources. The variable must contains a `,` separated list of key value `key=value`.

Fixes TP-496

### New feature (non-breaking change which adds functionality)

- Add a new environment variable `TOPOS_OTLP_TAGS` which will be used to define custom tags
	- `export TOPOS_OTLP_TAGS="environment=production,capabilities=superuser"`
- Tags are `,` separated and must have the `key=value` format

## PR Checklist:

### DIFF
--- a/crates/topos/src/main.rs
+++ b/crates/topos/src/main.rs
@@ -69,6 +69,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         ..
     }) = args.commands
     {
+        let resources = build_resources();
+
         global::set_text_map_propagator(TraceContextPropagator::new());
         let tracer = opentelemetry_otlp::new_pipeline()
             .tracing()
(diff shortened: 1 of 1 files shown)


crates/topos/src/main.rs
```diff
@@ -134,6 +129,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     }
 }
 
+fn build_resources() -> Vec<KeyValue> {
+    let mut resources = Vec::new();
+
+    resources.push(KeyValue::new(
+        "service.name",
+        std::env::var("TCE_JAEGER_SERVICE_NAME").expect("TCE_JAEGER_SERVICE_NAME must be defined"),
```
atanmarko, line 137: `TCE_JAEGER_SERVICE_NAME` is optional, you should not use expect here

atanmarko: Actually not to complicate things it will be handled in another PR #184

atanmarko (approved):

sebastiendan (approved):

crates/topos/src/main.rs
Freyskeyd: Yes I just reused the existing code

PR #185, opened 2023-03-17
feat: create checker for tce

Freyskeyd:
# Description

This PR is adding a new feature to the `topos` CLI which allow the push of a certificate in order to check the network. This feature is really not aiming to be used in production and should only be used in e2e. It is behind the `checker` feature flag.

## Additions and Changes

- Added a way to compare `uci::CertificateId` and `api::CertificateId`

## Fixes

Fix and improve reliability on `app_context` for the handling of `Echo/Ready` message when `Gossip` isn't arrived yet.

### New feature (non-breaking change which adds functionality)

- Created a new `checker` component that can be extended to other parts to check more things
- Create a checker for the `TCE` that:
	- Read a file in order to have the node list
	- Connect to every peer with `watch_certificates` listening for the subnet_id `[0u8;32]`
	- Connecting to a random peer to push the dummy certificate
	- Assert that all peers delivered it

## PR Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added or updated tests that comprehensively prove my change is effective or that my feature works

PR #192, opened 2023-03-22
feat: add GH actions to test push-certificate & local docker

Freyskeyd:
This PR aims to add some workflow to our CI in order to test some capabilities, such as spawning a local cluster and check that a certificate is delivered.

- Add `autoheal` docker container to local docker-compose in order to listen and reboot peers if they fails to sample due to concurrency.

## PR Checklist:


.github/workflows/checker.yml
```diff
@@ -0,0 +1,76 @@
...
+          username: ${{ env.PRIV_GH_USER }}
+          password: ${{ env.PRIV_GH_PACKAGE_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Building containers
+        run: |
+          docker-compose -f tools/docker-compose.yml build boot
+          docker-compose -f tools/docker-compose.yml up -d peer
```
sebastiendan, line 64: We're supposed to build but here we're running containers.

.github/workflows/checker.yml
```diff
@@ -0,0 +1,76 @@
...
+        target:
+          - { name: Linux, os: ubuntu-latest, triple: x86_64-unknown-linux-gnu }
+
+    name: Build and execute checker
+    runs-on: ${{ matrix.target.os }}
+    env:
+      PRIV_GH_USER: ${{ secrets.ROBOT_TOPOSWARE_USER }}
+      PRIV_GH_PACKAGE_TOKEN: ${{ secrets.ROBOT_TOPOSWARE_GH_PACKAGE_TOKEN }}
+    steps:
+      - uses: actions/checkout@v3
```
sebastiendan, line 50: Add a line-break

.github/workflows/checker.yml
sebastiendan: Add a line-break

.github/workflows/checker.yml
```diff
@@ -0,0 +1,76 @@
...
+          docker-compose -f tools/docker-compose.yml build boot
+          docker-compose -f tools/docker-compose.yml up -d peer
+      - name: Checking for healthy cluster
+        uses: nick-fields/retry@v2
+        with:
+          command: docker inspect --format "{{.State.Health.Status }}" $(docker-compose -f tools/docker-compose.yml ps -q peer)|grep "^healthy$"
+          retry_on: error
+          max_attempts: 3
+          retry_wait_seconds: 15
+          timeout_minutes: 5
```
sebastiendan, line 72: Add a line-break

.github/workflows/docker_utils.yml
```diff
@@ -75,9 +79,9 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=ref,event=branch,suffix=${{ inputs.features == 'tce-direct,sequencer' && '-direct' || ''}}
-            type=ref,event=tag,suffix=${{ inputs.features == 'tce-direct,sequencer' && '-direct' || ''}}
-            type=ref,event=pr,suffix=${{ inputs.features == 'tce-direct,sequencer' && '-direct' || ''}}
+            type=ref,event=branch,suffix=${{ inputs.docker_tag != '' && format('-{0}', input.docker_tag) || '' }}
```
sebastiendan, line 82: Instead of add yet another input, why don't we just set the suffix as the value of `features`? Then, it's clear what is behind an image and we prevent us from ending up with broken images (e.g., an image with `checker` docker tag but without the `checker` feature because of human error).

.github/workflows/checker.yml
```diff
@@ -0,0 +1,76 @@
+name: Checker
```
sebastiendan, line 1: I'm not a big fan of the `checker` word, doesn't feel natural. We need to be more explicit about what this workflow and the rust feature behind it are actually doing.

tools/docker-compose.yml
```diff
@@ -149,5 +151,14 @@ services:
       - RUST_LOG=info,topos=debug
       - TARGET_NODES_PATH=/tmp/shared/peer_nodes.json
 
+  autoheal:
```
sebastiendan, line 154: This is not documented. Can you add the rough list of changes as it is expected from the PR template?

sebastiendan (changes requested):

.github/workflows/checker.yml
Freyskeyd: There is already a linebreak. You mean a newline like this?

```suggestion
      - uses: actions/checkout@v3

      - name: Login to GitHub Container Registry
```

.github/workflows/checker.yml
Freyskeyd: > We're supposed to build but here we're running containers.

I can rename or create two steps indeed

.github/workflows/checker.yml
sebastiendan: Yes sorry, I meant newline.

.github/workflows/checker.yml
sebastiendan: Yes, I was implying changing also the feature itself, as checker doesn't feel explicit nor natural (in the sense of the English language)

Freyskeyd: Do you have any suggestion of another name?

.github/workflows/docker_utils.yml
Freyskeyd: If we want to use the `features` as a suffix, we need to parse the input and replace any `,` with `_` just for the tag.
Also, our base images will be:

- `topos-network/topos:tce-direct_sequencer`
- `topos-network/topos:tce_sequencer`
- `topos-network/topos:checker`

The more feature we will add the more long the tag will be, it also prevents nested feature activation but we can manage that. I'm ok to go with your approach but we need to think about what the end user will use.

Questions:
- How do we manage the root image if the suffix is always the `features` ?
- How do we manage feature changes over time (like rename, etc) ?

sebastiendan: > How do we manage the root image if the suffix is always the features ?
We can change the ternary condition to add the `${features}` suffix only if `features !== DEFAULT_FEATURES`.

> How do we manage feature changes over time (like rename, etc) ?
I guess a change in the name of a feature is as big on the native binary than it is on docker images, so I don't see a problem with it in the docker context.

.github/workflows/checker.yml
sebastiendan: If we're creating a certificate and passing it to a TCE network that we run just for that, it really resembles e2e tests to me (we have a debate on this with @hadjiszs :p), as we're using the API of the present repository and verify the rest of the code works as expected (i.e., delivery is working).

We're essentially doing the same thing on k8s, with cert-spammer.

>I want to be able to add more check to debug the broadcast

I don't think adding more checks is making this less e2e, probably the opposite.

sebastiendan: My comment was really about the feature as a whole. If we're not changing the feature name, no need to change the workflow.

>(if not misleading as not fully for e2e)

I'm starting to understand I'm missing something about this feature. Looking at the origin PR (#185—is it normal btw that the two PRs have the same title?), I read this:

>This feature is really not aiming to be used in production and should only be used in e2e.

We might want to update descriptions to make things consistent across PRs.

Can we just detail here which use cases for `checker` to be used are not e2e related?

.github/workflows/docker_utils.yml
```diff
@@ -75,9 +76,9 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=ref,event=branch,suffix=${{ inputs.features == 'tce-direct,sequencer' && '-direct' || ''}}
-            type=ref,event=tag,suffix=${{ inputs.features == 'tce-direct,sequencer' && '-direct' || ''}}
-            type=ref,event=pr,suffix=${{ inputs.features == 'tce-direct,sequencer' && '-direct' || ''}}
+            type=ref,event=branch,suffix=${{ inputs.features != env.DEFAULT_FEATURES && format('-{0}', input.features) || '' }}
```
sebastiendan, line 79: No problem with the formatting after all?

.github/workflows/test_tce_push_certificate.yml
```diff
@@ -0,0 +1,82 @@
...
+          registry: ghcr.io
+          username: ${{ env.PRIV_GH_USER }}
+          password: ${{ env.PRIV_GH_PACKAGE_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Building containers
+        run: |
+          docker-compose -f tools/docker-compose.yml build boot
```
sebastiendan, line 64: Is `docker compose` available instead of the deprecated `docker-compose`?

.github/workflows/test.yml
```diff
@@ -7,7 +7,52 @@ on:
...
+
 jobs:
+  cert_delivery:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 1
+      matrix:
+        value: ["first", "second", "third"]
+    steps:
+      - run: echo "running ${{ matrix.value }} cert_delivery"
```
hadjiszs, line 42: can't it be simply through the `--retries` of a single run of `cargo nextest`?

hadjiszs (approved):

.github/workflows/docker_utils.yml
Freyskeyd: reading [this](https://github.com/docker/metadata-action#image-name-and-tag-sanitization) I think we're good to go.

.github/workflows/test.yml
Freyskeyd: `retries` only run at least once, if the first run is a success then there is no more run. I want to be able to run the tests 3 times in a row and assert that the 3 run are successful

.github/workflows/docker_build_push.yml
```diff
@@ -38,3 +38,45 @@ jobs:
           ref: main
           wait_interval: 60
           client_payload: '{ "git_commit": "${{ github.sha }}", "docker_tag": "${{ env.docker_tag }}" }'
+
+  docker_e2e:
```
sebastiendan, line 42: One last comment: We could move `docker_e2e` before `k8s_e2e` and making the latter dependent on the former (with an extra `need`). This would save some CI time in case `docker_e2e` fails.

sebastiendan (approved):

.github/workflows/docker_build_push.yml
Freyskeyd: Ah yes good idea !

PR #197, opened 2023-03-24
fix(init.sh): Add exclusive lock for read/write with jq

foo0x29a:
# Description

- Split `jq` command into two
- Add safety check to include content only once

### Context

We seem to be experiencing some Byzantine Failures with our current stack during initialization. Sometimes launching our stack with Docker compose or K8s works fine, sometimes it fails with:

```
2023-03-24T13:39:55.576090Z DEBUG topos::components::tce: Start executing PushPeerList command
2023-03-24T13:39:55.577121Z DEBUG topos::components::tce: Executing the PushPeerList on the TCE service
Error: Custom { kind: UnexpectedEof, error: Error("EOF while parsing a value", line: 2, column: 0) }
```

This is very likely to be related with the file `peer_ids.json`.

Also, due to the ephemeral nature of containers/pods, restarts are part of their lifecycle. This causes duplicate content in the `peer_ids.json` and `peer_nodes.json` files.

### Hypothesis

The current stack is currently composed by:
- 01 boot node
- 15 peer nodes

Launching them with Docker compose or in a K8s cluster, containers are created (almost) **simultaneously** .
The following line of code in especial will be the subject of this PR:

```
cat <<< $($JQ --arg PEER $PEER '. += [$PEER]' $PEER_LIST_PATH) > $PEER_LIST_PATH
```

The `jq` command **doesn't** provide in-place editing feature like tools such as `sed`. One alternative is the above solution, where:
- `$PEER_LIST_PATH` is read in a command substitution
- output is redirected to `$PEER_LIST_PATH`

During the `stdout` redirection, the `$PEER_LIST_PATH` file is [truncated to zero-length](https://www.gnu.org/software/bash/manual/bash.html#Redirecting-Output) and only then content is written.

The concurrency between containers/pods trying to read/write to the same file may cause a situation where `container A` reads `$PEER_LIST_PATH` while `container B` truncates `$PEER_LIST_PATH` .

### Proposal
The proposal here is to instead:

```
           # Acquire lock and add $PEER to ${PEER_LIST_PATH} only once
           (
               flock --exclusive 200
               cat <<< $($JQ --arg PEER $PEER '. += [$PEER]|unique' $PEER_LIST_PATH) > $PEER_LIST_PATH

           ) 200>/tmp/shared/"${PEER_LIST_PATH}.lock"
```

1. Redirect stdout of subshell to `"${PEER_LIST_PATH}.lock"` through fd `200`
2. Acquire exclusive lock for fd `200` (inherited from parent process)
3. Use `unique` to add content only once

## PR Checklist:

### DIFF
--- a/tools/init.sh
+++ b/tools/init.sh
@@ -63,7 +63,7 @@ case "$1" in
            # Acquire lock and add $PEER to ${PEER_LIST_PATH} only once
            (
                flock --exclusive -w 10 201 || exit 1
-               cat <<< $($JQ --arg PEER $PEER '. += [$PEER]|unique' $PEER_LIST_PATH) > $PEER_LIST_PATH
+               cat <<< $($JQ --arg PEER $PEER '. |= (. + [$PEER] | unique )' $PEER_LIST_PATH) > $PEER_LIST_PATH
 
            ) 201>"${PEER_LIST_PATH}.lock"
 


sebastiendan (approved):
Awesome description of the problem and its solution, great work @foo0x29a! 🔥

tools/init.sh
```diff
@@ -59,7 +59,13 @@ case "$1" in
            done
 
            PEER=$($TOPOS_BIN tce keys --from-seed=$HOSTNAME)
-           cat <<< $($JQ --arg PEER $PEER '. += [$PEER]' $PEER_LIST_PATH) > $PEER_LIST_PATH
+
+           # Acquire lock and add $PEER to ${PEER_LIST_PATH} only once
+           (
+               flock --exclusive -w 10 201 || exit 1
```
Freyskeyd, line 65: Great ! I was wondering if the `fd` was shared across container with shared volume, but I think it is.
One way to test the lock would be to add a `sleep` of 15 sec for one of the container just after the lock is acquired. That way, we should see the failure on other nodes.

PR #200, opened 2023-03-28
feat: get pending certificates

atanmarko:
# Description

Add tce api function `GetLastPendingCertificates` to retrieve latest generated pending certificate for list of subnet ids.
Sequencer on restart now continues generating certificates from latest previously generated and pushed certificate to the TCE 

Fixes TP-508

## Additions and Changes

Add api function primarily for sequencer to determine latest certificate it produced so that it could continue generating certificates from that point.
Updated tce storage api to implement this functinality.

## Breaking changes

TCE GRPC api is changed, client needs to be rebuilt.

## PR Checklist:


Freyskeyd (changes requested):

crates/topos-api/proto/topos/tce/v1/api.proto
```diff
@@ -32,6 +36,18 @@ message GetSourceHeadResponse {
   topos.uci.v1.Certificate certificate = 2;
 }
 
+message GetLastPendingCertificatesRequest {
+  repeated topos.shared.v1.SubnetId subnet_ids = 1;
+}
+
+message LastPendingCertificate {
```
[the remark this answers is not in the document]
atanmarko: Good idea

crates/topos-sequencer-tce-proxy/src/lib.rs
```diff
@@ -146,6 +156,23 @@ impl TceClient {
...
+
+    pub async fn get_last_pending_certificates(
+        &mut self,
+        subnet_ids: Vec<SubnetId>,
+    ) -> Result<HashMap<SubnetId, Option<Certificate>>, Error> {
+        #[allow(clippy::type_complexity)]
+        let (sender, receiver): (
+            oneshot::Sender<Result<HashMap<SubnetId, Option<Certificate>>, Error>>,
+            oneshot::Receiver<Result<HashMap<SubnetId, Option<Certificate>>, Error>>,
+        ) = oneshot::channel::<Result<HashMap<SubnetId, Option<Certificate>>, Error>>();
```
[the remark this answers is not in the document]
atanmarko: Right I forgot to remove it when I added the type to oneshot::channel::<>

crates/topos-api/src/conversions/shared/v1/certificate.rs
```diff
@@ -2,7 +2,7 @@ use super::v1::CertificateId;
 
 impl std::fmt::Display for CertificateId {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", hex::encode(&self.value))
+        write!(f, "0x{}", hex::encode(&self.value))
```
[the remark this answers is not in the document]
atanmarko: We can switch to binary `base64` but then debug printout is not usable from the context of our APP, because GRPC encoding/decoding is transparent to us (unless we are debugging/fixing GRPC protocol itself).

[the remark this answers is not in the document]
atanmarko: Ok, lets then switch to base64 for Display of protobuf generated structures

crates/topos-tce-storage/src/tests/mod.rs
```diff
@@ -368,13 +374,71 @@ async fn get_source_head_for_subnet(storage: RocksDBStorage) {
...
+    let certificates_for_source_subnet_2 =
+        create_certificate_chain(SOURCE_SUBNET_ID_2, &[TARGET_SUBNET_ID_2], 15);
+    for cert in &certificates_for_source_subnet_2[0..10] {
+        storage.persist(cert, None).await.unwrap();
+    }
+    for cert in &certificates_for_source_subnet_2[10..] {
+        storage.add_pending_certificate(cert).await.unwrap();
+        expected_pending_certificates.push((expected_pending_certificates_count, cert.clone()));
+        expected_pending_certificates_count += 1;
+    }
```
hadjiszs, line 414:
```suggestion
    let certificates_for_source_subnet_1 =
        create_certificate_chain(SOURCE_SUBNET_ID_1, &[TARGET_SUBNET_ID_2], 15);
    let certificates_for_source_subnet_2 =
        create_certificate_chain(SOURCE_SUBNET_ID_2, &[TARGET_SUBNET_ID_2], 15); 
  
    // Persist the first 10 Cert of each Subnets
    for cert in &certificates_for_source_subnet_1[0..10] {
        storage.persist(cert, None).await.unwrap();
    }
    for cert in &certificates_for_source_subnet_2[0..10] {
        storage.persist(cert, None).await.unwrap();
    }
    
    // Add the last 5 cert of each Subnet as pending certificate
    for cert in &certificates_for_source_subnet_1[10..] {
        storage.add_pending_certificate(cert).await.unwrap();
        expected_pending_certificates.push((expected_pending_certificates_count, cert.clone()));
        expected_pending_certificates_count += 1;
    }
    for cert in &certificates_for_source_subnet_2[10..] {
        storage.add_pending_certificate(cert).await.unwrap();
        expected_pending_certificates.push((expected_pending_certificates_count, cert.clone()));
        expected_pending_certificates_count += 1;
    }
```

suggestion to make the intention clearer, at least to add 1~2 comments to understand quicker what's the  setup

crates/topos-tce/src/app_context.rs
```diff
@@ -209,6 +209,41 @@ impl AppContext {
...
+                        .iter()
+                        .map(|subnet_id| (subnet_id.to_string(), None))
+                        .collect();
+
+                if let Ok(pending_certificates) =
+                    self.pending_storage.get_pending_certificates().await
+                {
+                    // Iterate through pending certificates and determine last one for every subnet
+                    // Last certificate in the subnet should be one with the highest index
+                    for (pending_certificate_id, cert) in pending_certificates.into_iter().rev() {
```
hadjiszs, line 228: we probably want to rely on the field `position` of the cert rather than on the order of the column, since the column is inserting for a FIFO access I think, so normally nothing prevents one peer to insert an older cert after a more recent one (in terms of position). We could have it as a separate ticket if it has indeed to be done :ok_hand:

crates/topos-tce-storage/src/connection/handlers.rs
```diff
@@ -70,6 +72,39 @@ where
...
+            self.storage.get_pending_certificates().await.map_err(|e| {
+                error!("Failure on the storage to get pending certificates: {e}");
+                e
+            })?;
+
+        pending_certificates.sort_by(|(pending_cert_id_a, _), (pending_cert_id_b, _)| {
+            pending_cert_id_a
+                .partial_cmp(pending_cert_id_b)
+                .unwrap_or(Ordering::Equal)
+        });
```
Freyskeyd, line 102: I don't think you need to sort them, they are already sorted, I guess.

crates/topos-api/src/generated/topos.tce.v1.rs
```diff
@@ -325,6 +375,18 @@ pub mod api_service_server {
             &self,
             request: tonic::Request<super::GetSourceHeadRequest>,
         ) -> Result<tonic::Response<super::GetSourceHeadResponse>, tonic::Status>;
+        /// / This RPC allows a client to get latest pending certificates for
```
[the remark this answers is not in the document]
atanmarko: @hadjiszs This is generated code, seems Rust protobuf compiler puts comment on both client and server function code

crates/topos-tce/src/app_context.rs
Freyskeyd: There is no `position` on a `pending_certificates` and we can't enforce ordering before the broadcast happens. There is no clear solution to do what you suggest. The whole `get last pending` features is regarding a TCE node state, not a sequencer. It's up to the sequencer to decide whether it can trust this TCE or not.

crates/topos-sequencer-tce-proxy/tests/tce_tests.rs
```diff
@@ -231,6 +232,129 @@ async fn test_tce_get_source_head_certificate(
...
+async fn test_tce_get_last_pending_certificates(
+    #[future] start_node: TceContext,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let mut context = start_node.await;
+
+    let source_subnet_id: SubnetId = SOURCE_SUBNET_ID_1.into();
+    let target_subnet_id: SubnetId = TARGET_SUBNET_ID_1.into();
+    let prev_certificate_id: CertificateId = PREV_CERTIFICATE_ID.into();
+    let certificate_id_1: CertificateId = CERTIFICATE_ID_1.into();
+    let certificate_id_2: CertificateId = CERTIFICATE_ID_2.into();
```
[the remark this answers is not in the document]
atanmarko: Right, I will refactor this test

crates/topos-tce-api/src/grpc/mod.rs
```diff
@@ -179,19 +183,71 @@ impl ApiService for TceGrpcService {
...
+        }
+
+        receiver
+            .map(|value| match value {
+                Ok(Ok(map)) => Ok(Response::new(GetLastPendingCertificatesResponse {
+                    last_pending_certificate: map
+                        .into_iter()
+                        .map(|(subnet_id, cert)| {
+                            (
+                                subnet_id,
```
Freyskeyd, line 233: You need to `base64` encode here following the discussion we had

crates/topos-tce/src/app_context.rs
atanmarko: @hadjiszs Position is defined when you deliver the certificate - as implemented now, pending certificates do not have position. We can discuss this topic when we further refactor pending certificates handling

Freyskeyd (approved):

crates/topos-tce-storage/src/connection/handlers.rs
atanmarko: They are, but just in case and no harm is done

Freyskeyd: It just adds runtime overhead to go through the whole list

crates/topos-tce-api/src/grpc/mod.rs
atanmarko: I don't think it applies here - this is not about debug display. It is about result - and we can not have `SubnetId` as key in the map so we have to use string representation. Both sequencer and tce are handling SubnetId as hex encoded string when dealing with them so I would keep hex encoded string here.

atanmarko: If you use base64 encoded string here it will be again base64 encoded when grpc performs encoding/decoding

crates/topos-tce-storage/src/connection/handlers.rs
atanmarko: It adds microsecond or two to couple of hundred of milliseconds http call

atanmarko: As we will refactor handling of the pending certificates anyway better for sorting to stay here to communicate what we need

crates/topos-api/src/generated/topos.tce.v1.rs
```diff
@@ -283,6 +306,33 @@ pub mod api_service_client {
             );
             self.inner.unary(request.into_request(), path, codec).await
         }
+        /// / This RPC allows a client to get latest pending certificates for
+        /// / requested subnets (by their subnet id)
+        /// /
+        /// / Returns a map of subnet_id -> last pending certificate
+        /// / If there are no pending certificate for a subnet, returns None for that subnet id
```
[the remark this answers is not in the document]
atanmarko: @hadjiszs This is generated code, seems Rust protobuf compiler puts comment from proto file on both client and server function code

crates/topos-api/src/generated/topos.tce.v1.rs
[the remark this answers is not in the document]
atanmarko: I accidentally deleted duplicated @Freyskeyd comment but when I refreshed page both were gone. His comment was similar to my note from above

crates/topos-tce/src/app_context.rs
hadjiszs: Right, I mentioned about the field `position` but should refer to the field `prev_id` instead. The ordering of a source stream is already defined prior to the broadcast, thanks to the `prev_id`. We are not talking about the ordering on the point of view of a target subnet here.

Given the source subnet A

```
Cert 0x01 {
  source: subnet_a,
  prev_id: 0x00
}

Cert 0x02 {
  source: subnet_a,
  prev_id: 0x01
}
```

Here we could have 0x01 after 0x02 in our pending column, and we'll send `0x01` as the last pending for `subnet_a`, while we should send `0x02` instead. And we have the data to know that the latest one is 0x02 thanks to the `prev_id`, we don't need to proceed to the broadcast to know that.

The problem happens if there is holes in our pending column, e.g. we have `0x00` and `0x02` but not `0x01`, then we can't know the ordering between `0x00` and `0x02`, so in that case it sounds legit to take the last one in the column.

Also, it's not about trust issues here, just about the data itself that is returned and tagged as "last pending", which is not in the point of view of the TCE node itself if we rely only on the order of the column of the pending.

crates/topos-tce-api/src/grpc/mod.rs
Freyskeyd: In the `get_last_pending_certificates` response, the type of the key is `string` not `bytes` so there is no `base64` encoding by default. As example:

```
{
  "subnetIds": [
    {
      "value": "MTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTE="
    },
    {
      "value": "MTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTI="
    }
  ]
}
```

This is a valid payload for the query with the `subnet_ids: [1111111111111111111111111111111, 11111111111111111111111111111112];`.

Here's the response with a dummy certificate on one subnet:

```
{
  "lastPendingCertificate": {
    "0x3131313131313131313131313131313131313131313131313131313131313131": {
      "value": {
        "prevId": {
          "value": "MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA="
        },
        "sourceSubnetId": {
          "value": "MTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTE="
        },
        "stateRoot": "MTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTE=",
        "txRootHash": "MTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTE=",
        "targetSubnets": [
          {
            "value": "MTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTE="
          }
        ],
        "verifier": 507,
        "id": {
          "value": "NTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTE="
        },
        "proof": {
          "value": "MTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTE="
        },
        "signature": {
          "value": "MTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTE="
        }
      }
    },
    "0x3131313131313131313131313131313131313131313131313131313131313132": {}
  }
}
```

As you can see, the `key` isn't encoded in base64, the `sourceSubnetid` is encoded tho. The main issue for me is that we have a lack of consistency here.

crates/topos-tce/src/app_context.rs
Freyskeyd: @hadjiszs this is a known issue and the main problem is the one you're describing. I think that we'll need to consume and requeue the certificates that fall into this hole.

crates/topos-tce-api/src/grpc/mod.rs
atanmarko: Ok, I will change encoding to `base64` here

PR #202, opened 2023-03-30
test: add a new e2e file to github actions

Freyskeyd:
# Description

This PR adds a new github action file to handle the different e2e triggers that we want to execute.
The goal is to execute the e2e after the docker build and push in order to add tests that use those images and doesn't rebuild the entire project locally.

The proper test run will be added in another PR.

## PR Checklist:

### DIFF
--- a/.github/workflows/docker_build_push.yml
+++ b/.github/workflows/docker_build_push.yml
@@ -1,4 +1,4 @@
-name: Docker_build_push
+name: Docker build and push
 
 on:
   push:
--- a/.github/workflows/docker_utils.yml
+++ b/.github/workflows/docker_utils.yml
@@ -1,4 +1,4 @@
-name: Docker build and push
+name: template - docker
 
 env:
   REGISTRY: ghcr.io
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,49 @@
+name: Execute E2E tests
+
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows: ["Docker build and push"]
+    branches: ["main"]
+    types:
+      - completed
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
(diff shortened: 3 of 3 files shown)


sebastiendan:
@Freyskeyd How do you plan on connecting this new workflow with the actual tests? I'm not sure to understand what we're doing here with pulling and simply tagging images.

As to the trigger, with @foo0x29a we had seen issues with workflows triggered once other are completed (with this exact pattern). The other logic that Thiago went with is having the docker job written in the `util` workflow and having different higher level workflows use it and run what needs to be run afterwards.

You have an example [here](https://github.com/topos-network/topos/blob/main/.github/workflows/docker_build_push.yml#L20) with `k8s-e2e`. Actually intuitively I would suggest having your new test job in `docker_build_push` as well, right before the `k8s-e2` job, so that we run the following in order, and only when each step completes successfully:

```
- docker
- docker compose e2e
- k8s e2e
```

Freyskeyd:
> @Freyskeyd How do you plan on connecting this new workflow with the actual tests? I'm not sure to understand what we're doing here with pulling and simply tagging images.
 
As described in the description, the test will come later. The test part is defined in the PR #192

> As to the trigger, with @foo0x29a we had seen issues with workflows triggered once other are completed (with this exact pattern). The other logic that Thiago went with is having the docker job written in the `util` workflow and having different higher level workflows use it and run what needs to be run afterwards.

Mmh that's pretty weird because I used this technics a lot in previous projects/companies. What kind of issues did you encounter?

The goal here is to really have one workflow triggered only when the docker image is built.

It's cleaner to test the workflow of our local setup (which use docker container) using the generated image than rebuilding it from scratch. That way we're asserting that the pushed container that people will pull is check and verified.

To build this, I'm pulling the image pushed by the previous workflow and I'm tagging this image locally to `main` .

**Why:** The `docker-compose.yml` is using this tag by default, I didn't find any clean solution to replace the tag in the yml file so I'm tagging the image as `main` in order to prevent `docker compose` to pull it when we execute the `up` command.

PR #229, opened 2023-05-11
ci: add new grpc js client stubs generation CI workflow

sebastiendan:
# Description

This PR adds JavaScript client stub generation and packaging in a [@topos-network/topos-grpc-client-stub](https://www.npmjs.com/package/@topos-network/topos-grpc-client-stub) npm package.

Client stubs can then be used in web applications for interactions with the gRPC API of a `topos` node.

## Additions and Changes

- Add `package.json` and `package-lock.json` files in the `proto` directory to identify and version the npm package
- Add a CI workflow to generate stubs and package them in a new package to be published

**Note: Packages are built and published on new Github `releases`. It's important to remember that the version of the package (specified in its `package.json` file) needs to be updated priori to the creation of the release on Github (otherwise package publication will fail and the release will have to be recreated).**

Tagging @Freyskeyd @atanmarko @hadjiszs @gruberb to double-check this point above ⬆️ 

## PR Checklist:

- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added or updated tests that comprehensively prove my change is effective or that my feature works

PR #232, opened 2023-05-12
ci: generate ts typings for grpc client stub

sebastiendan:
# Description

#229 introduced JavaScript gRPC client stubs generation through a new CI workflow. This PR adds the generation of TypeScript typings so that TypeScript clients can better use our client stubs.

Additionally, this PR changes the npm package publication flow to the following one:
- First of all, the CI workflow will only run when
  - the push on `main` contains changes to protobuf files
  - the PR contains changes to protobuf files (coming from the active update or previous ones)
- On PR updates, a step will check whether the npm package version was bumped and make the job fail if not
- On merge on `main`, the `npm publish` step is run if the above was successful

Testing this new logic:
- [run](https://github.com/topos-network/topos/actions/runs/5072802220) that shows a failure because of missing version bump
- [run](https://github.com/topos-network/topos/actions/runs/5072886924/jobs/9111190112#step:9:20) that shows a success because version was bumped

## Additions and Changes

- Add the `grpc_tools_node_protoc_ts` npm package as dep
- Install `protoc` in the CI workflow
- Use `protoc` with a bin packaged in `grpc_tools_node_protoc_ts` to generate typings
- Bump npm package version to `0.1.1`

## PR Checklist:

### DIFF
--- a/.github/workflows/grpc-client-stubs.yml
+++ b/.github/workflows/grpc-client-stubs.yml
@@ -58,6 +58,7 @@ jobs:
 
       - name: Check if npm package version has been updated
         id: check
+        working-directory: ./crates/topos-api/proto
         uses: EndBug/version-check@v2
 
       - name: Publish npm package
--- a/.github/workflows/grpc-client-stubs.yml
+++ b/.github/workflows/grpc-client-stubs.yml
@@ -1,8 +1,14 @@
 name: gRPC client stubs generation
 
 on:
-  release:
-    types: [published]
+  push:
+    branches: [main]
+    paths:
+      - "**.proto"
+  pull_request:
+    branches: [main]
+    paths:
+      - "**.proto"
 
 jobs:
   js-grpc:
@@ -50,7 +56,12 @@ jobs:
         run: >
           protoc --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts --ts_out=grpc_js:./generated -I . $(find ./topos -name '*.proto')
 
+      - name: Check if npm package version has been updated
+        id: check
+        uses: EndBug/version-check@v2
+
       - name: Publish npm package
+        if: steps.check.outputs.changed == 'true' && ${{ github.event_name == 'push' }} && ${{ github.ref == 'refs/heads/main' }}
         working-directory: ./crates/topos-api/proto
         run: npm publish --access public
         env:
--- a/.github/workflows/grpc-client-stubs.yml
+++ b/.github/workflows/grpc-client-stubs.yml
@@ -58,8 +58,9 @@ jobs:
 
       - name: Check if npm package version has been updated
         id: check
-        working-directory: ./crates/topos-api/proto
         uses: EndBug/version-check@v2
+        with:
+          file-name: ./crates/topos-api/proto/package.json
 
       - name: Publish npm package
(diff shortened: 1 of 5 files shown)


.github/workflows/grpc-client-stubs.yml
```diff
@@ -38,6 +45,20 @@ jobs:
...
+          protoc --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts --ts_out=grpc_js:./generated topos/shared/v1/frost.proto &&
+          protoc --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts --ts_out=grpc_js:./generated topos/shared/v1/stark_proof.proto &&
+          protoc --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts --ts_out=grpc_js:./generated topos/shared/v1/subnet.proto &&
+          protoc --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts --ts_out=grpc_js:./generated topos/shared/v1/uuid.proto &&
+          protoc --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts --ts_out=grpc_js:./generated topos/uci/v1/certification.proto &&
+          protoc --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts --ts_out=grpc_js:./generated topos/tce/v1/api.proto &&
+          protoc --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts --ts_out=grpc_js:./generated topos/tce/v1/console.proto &&
+          protoc --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts --ts_out=grpc_js:./generated topos/tce/v1/synchronization.proto
```
Freyskeyd, line 60: You can use something like that:

```suggestion
        run: >
          protoc --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts --ts_out=grpc_js:./generated -I . $(find ./topos -name '*.proto')
```

Freyskeyd (approved):
Could you rename the github action workflow to `gRPC client stubs generation` as we could add more language to this workflow at some point.

gruberb (approved):
LGTM 👍

gruberb (commented):
Looks good, thanks for the detailed write-up! 

Just wondering if there is a scenario where the types of the JS gRPC client could be newer than the current release? Merging with `main` doesn't mean release, so the first logic seem to make more sense to me (because now we could have a situation with unreleased, but updated types in `main` the JS client is using, but the currently deployed `topos` network is not yet using).

gruberb (commented):
Looks good, thanks for the detailed write-up! 

Just wondering if there is a scenario where the types of the JS gRPC client could be newer than the current release? Merging with `main` doesn't mean release, so the first logic seem to make more sense to me (because now we could have a situation with unreleased, but updated types in `main` the JS client is using, but the currently deployed `topos` network is not yet using).

sebastiendan:
> Looks good, thanks for the detailed write-up!
> 
> Just wondering if there is a scenario where the types of the JS gRPC client could be newer than the current release? Merging with `main` doesn't mean release, so the first logic seem to make more sense to me (because now we could have a situation with unreleased, but updated types in `main` the JS client is using, but the currently deployed `topos` network is not yet using).

Housekeeping: This was replied on [Slack](https://toposware.slack.com/archives/C045B9XKRFE/p1685036136069189?thread_ts=1684953988.734629&cid=C045B9XKRFE)

PR #233, opened 2023-05-15
chore: improve spans to have benchmarks

Freyskeyd:
# Description

This PR aims to improve and test the aws benchmarks. Many things have been done in order to have a working benchmarks:

- [x] Remove the sampler
- [x] Remove the use of Request/Response for sampling
- [x] Remove the use of Request/Response for broadcast
- [x] Switch to GossipSub
- [x] Rework the spammer to have more predictability (batches, gracefull stop)
- [x] Upgrade to last version of libp2p
- [x] Expose `/metrics` about the broadcast and gossip
- [x] Rework the double echo to better handle huge amount of certificates.

Fixes TP-609
Fixes TP-517

## PR Checklist:


atanmarko:
Hey @Freyskeyd  what is the purpose of this PR, is it still work in progress as you requested review?

crates/topos-tce/src/app_context.rs
```diff
@@ -267,315 +262,319 @@ impl AppContext {
...
-                spawn(async move {
-                    let _results = join_all(future_pool)
-                        .instrument(span.clone())
-                        .with_context(span.context().clone())
-                        .await;
                 });
+
+                if let Err(e) = self
+                    .network_client
+                    .publish::<NetworkMessage>(topos_p2p::TOPOS_ECHO, data)
```
hadjiszs, line 551:
```suggestion
                    .publish::<NetworkMessage>(topos_p2p::TOPOS_READY, data)
```

looks like that brings an infinite loop of messages as from the doc the `message_id` is by default `( peer_id | sequence_number )` of the publisher, so it seems to me that they will be taken as different payloads if I'm not mistaken

btw we may want then to switch the default [message_id()] implementation (https://docs.rs/gossipsub/latest/gossipsub/struct.GossipsubConfig.html#method.message_id) from the concatenation to the hash of the message instead no?

crates/topos-tce-broadcast/src/double_echo/mod.rs
```diff
@@ -276,52 +507,79 @@ impl DoubleEcho {
     }
 
     #[cfg_attr(feature = "direct", allow(dead_code))]
-    pub(crate) fn handle_broadcast(&mut self, cert: Certificate) {
+    pub(crate) fn handle_broadcast(&mut self, cert: Certificate, origin: bool) {
```
hadjiszs, line 510: Not sure to get the purpose of this `origin`, I found it set only on the tests to true

Is it somewhat to track whether the certificate is coming from the API or from gossip or something else?

Freyskeyd: Yep it's to define if the we need to gossip the certificate when receiving the cert via API

gruberb (changes requested):

hadjiszs (approved):

crates/topos-tce/src/lib.rs
```diff
@@ -48,17 +50,46 @@ pub async fn run(
...
-    info!("I am {}", peer_id);
+    warn!("I am {}", peer_id);
     tracing::Span::current().record("peer_id", &peer_id.to_string());
 
     let external_addr: Multiaddr =
         format!("{}/tcp/{}", config.tce_addr, config.tce_local_port).parse()?;
 
     let addr: Multiaddr = format!("/ip4/0.0.0.0/tcp/{}", config.tce_local_port).parse()?;
 
+    use axum::{routing::get, Router};
```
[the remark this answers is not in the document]
Freyskeyd: I didn't wanted to add extract content to this PR. This part will need to be refactor into another crate yes.

gruberb (approved):
Needs to be refactored a bit, but can agree that we need this in to move forward 👍

PR #234, opened 2023-05-15
feat: send token sequencer test

atanmarko:
# Description

Implemented sequencer test with send tocken and certificate cration

Fixes TP-594

## PR Checklist:

### DIFF
--- a/crates/topos-sequencer-subnet-client/src/lib.rs
+++ b/crates/topos-sequencer-subnet-client/src/lib.rs
@@ -257,8 +257,11 @@ impl SubnetClient {
                 .parse()
                 .map_err(Error::WalletError)?
         } else {
-            // Dummy key, can not sign transactions
-            hex::encode([0u8; 32]).parse().map_err(Error::WalletError)?
+            // Dummy random key, can not sign transactions
+            // Could not be zero key
+            "bbb92d6e98884f76de468fa3f6278f8807c48bebc13595d45af5bdc4da7021aa"
+                .parse()
+                .map_err(Error::WalletError)?
         };
         let chain_id = http
             .get_chainid()
--- a/crates/topos-sequencer-subnet-client/src/lib.rs
+++ b/crates/topos-sequencer-subnet-client/src/lib.rs
@@ -174,7 +174,6 @@ impl SubnetClientListener {
             .await
             .map_err(Error::EthersProviderError)?
             .ok_or(Error::InvalidBlockNumber(next_block_number))?;
-
         let block_number = block
             .number
             .ok_or(Error::InvalidBlockNumber(next_block_number))?;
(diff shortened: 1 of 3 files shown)


crates/topos-sequencer-subnet-client/src/lib.rs
```diff
@@ -258,8 +257,11 @@ impl SubnetClient {
                 .parse()
                 .map_err(Error::WalletError)?
         } else {
-            // Dummy key, can not sign transactions
-            hex::encode([0u8; 32]).parse().map_err(Error::WalletError)?
+            // Dummy random key, can not sign transactions
+            // Could not be zero key
+            "bbb92d6e98884f76de468fa3f6278f8807c48bebc13595d45af5bdc4da7021aa"
+                .parse()
+                .map_err(Error::WalletError)?
```
gruberb, line 264: Seems overkill, but how about we do it randomly?

```rust    
let mut rng = rand::thread_rng();
let mut bytes = [0u8; 32];
rng.fill(&mut bytes[..]);

hex::encode(bytes);
```

Or we have a constant? I also worry about hardcoded "magic strings", even if they are obvious and commented! 

```rust
const DUMMY_MAGIC_KEY: &'static str = "bbb92d6e98884f76de468fa3f6278f8807c48bebc13595d45af5bdc4da7021aa"
```

crates/topos-sequencer-subnet-runtime/tests/subnet_contract.rs
```diff
@@ -261,6 +278,70 @@ async fn deploy_contracts(
...
+    ));
+
+    let i_topos_messaging = IToposMessaging::new(topos_messaging_address, client.clone());
+
+    // Deploy token
+    let token_name: Token = Token::String("Test Token".to_string());
+    let token_symbol: Token = Token::String("TKX".to_string());
+    let token_mint_cap: Token = Token::Uint(U256::from(100_000_000));
+    let token_address_zero: Token =
+        Token::Address("0000000000000000000000000000000000000000".parse()?);
```
gruberb, line 302: Should we have a constant here instead of hardcoding it?

crates/topos-sequencer-subnet-runtime/tests/subnet_contract.rs
```diff
@@ -625,3 +706,113 @@ async fn test_subnet_id_call(
...
+    if let Err(e) = context
+        .i_topos_messaging
+        .send_token(
+            TARGET_SUBNET_ID_2.into(),
+            "00000000000000000000000000000000000000AA".parse()?,
+            i_erc20.address(),
+            U256::from(2),
+        )
+        .legacy()
+        .gas(5_000_000)
```
gruberb, line 779: Same for the gas variables. If we have these as "parameter variables" at the beginning, it's easier to reason and understand?

gruberb (commented):
Nothing too major, just suggesting adding `cont`s and variables instead of directly adding `&str`.

gruberb (approved):
LGTM 👍

PR #237, opened 2023-05-29
chore: update subnet contract abi retrieval

atanmarko:
# Description

Get latest smart contract ABI from latest npm smart contract release

Fixes TP-501

## PR Checklist:


crates/topos-sequencer-subnet-runtime/tests/subnet_contract.rs
```diff
@@ -46,12 +46,12 @@ const CERTIFICATE_ID_3: CertificateId = CERTIFICATE_ID_8;
...
-abigen!(IToposCore, "npm:@topos-network/topos-smart-contracts@1.0.1/artifacts/contracts/interfaces/IToposCore.sol/IToposCore.json");
-abigen!(IToposMessaging, "npm:@topos-network/topos-smart-contracts@1.0.1/artifacts/contracts/interfaces/IToposMessaging.sol/IToposMessaging.json");
+abigen!(TokenDeployerContract, "npm:@topos-network/topos-smart-contracts@latest/artifacts/contracts/topos-core/TokenDeployer.sol/TokenDeployer.json");
+abigen!(ToposCoreContract, "npm:@topos-network/topos-smart-contracts@latest/artifacts/contracts/topos-core/ToposCore.sol/ToposCore.json");
+abigen!(ToposCoreProxyContract, "npm:@topos-network/topos-smart-contracts@latest/artifacts/contracts/topos-core/ToposCoreProxy.sol/ToposCoreProxy.json");
+abigen!(ToposMessagingContract, "npm:@topos-network/topos-smart-contracts@latest/artifacts/contracts/topos-core/ToposMessaging.sol/ToposMessaging.json");
+abigen!(IToposCore, "npm:@topos-network/topos-smart-contracts@latest/artifacts/contracts/interfaces/IToposCore.sol/IToposCore.json");
+abigen!(IToposMessaging, "npm:@topos-network/topos-smart-contracts@latest/artifacts/contracts/interfaces/IToposMessaging.sol/IToposMessaging.json");
```
hadjiszs, line 54: shouldn't we specify an explicit version and manually bump it with companion PR whenever we get breaking changes on topos-smart-contract?

atanmarko: @hadjiszs CC @sebastiendan @JDawg287 I think latest release version is less error prone. When we come to the point to have multiple incompatible releases of the system then maybe manually specifying version is ok. 

Also problem with `abigen!` is that it is macro very allergic to any non string literal and I did not find a way to concatenate a string to pass version, so I had to manually type version for all of the contracts

hadjiszs: I guess that's not so much an issue right now but I believe that using explicit version will be easier, e.g., while developing on the sequencer side, I believe that you'd not want to get breaking changes after recompiling due to an update of `latest`, then locally you'll probably explicitly switch to the previous version to develop the feature

atanmarko: @hadjiszs I would say purpose of these tests is to find those issues and incompatibilities, so that they are immediately fixed. Sequencer master branch should follow smart contract master branch

hadjiszs: I agree for the test side that then the only consequence is that the feature PR will be prevented from being merged because of failing CI, until there is the fix, which is fine by me.

Just to be sure regarding the reference to latest in the `crates/topos-sequencer-subnet-client/src/subnet_contract.rs`, doesn't it prevent from even compiling and running locally on the feature branch as soon as there is a breaking change updated on `topos-smart-contracts@latest`? As long as the update to comply with topos-smart-contracts is there?
In which case you'll want either to make the fix beforehand in parallel to integrate it, or you'll refer to a specific previous version instead of latest no?

atanmarko: @hadjiszs Right

hadjiszs (approved):

PR #239, opened 2023-06-05
feat: add graphql server and endpoints to a topos node

gruberb:
# Description

This PR adds a GraphQL server to a topos node. It exposes two endpoints on port `4000`:

* `GET`: GraphQL playground
* `POST`: Open GraphQL endpoint to send queries/mutations/subscriptions

So far there is one possible query: `certificates`.

```graphql
query {
  certificates(
    fromSourceCheckpoint: {
      sourceSubnetIds: [
        { value: "0xHEX_VALUE" }
      ],
      positions: [
        {
          sourceSubnetId: { value: "0xHEX_VALUE" },
          position: 0,
        }
      ]
    },
    first: 10
  ) {
    id
    prevId
    proof
    signature
    sourceSubnetId
    stateRoot
    targetSubnets {
      value
    }
    txRootHash
    verifier
  }
}
```

Fixes [TP-617](https://linear.app/toposware/issue/TP-617/as-topos-we-want-to-expose-the-delivered-certificates-through-the-api). 

## Additions and Changes

 * I added features flags for `topos-api`: `grpc` and `graphql`. The `topos-api` is now exposing both features through a higher level `lib.rs` file, and the files are located in their respectives folders inside the `src` folder.
 * Two added `ENV` variables: `TCE_GRAPHQL_API_ADDR=0.0.0.0:4000` and port `4000`
 * The `boot` node is exposing the new port: `4000:4000`

### New feature (non-breaking change which adds functionality)

Upon start of a node, you can access a GraphQL playground on port `4000`:

Screenshot 2023-06-13 at 12 10 45

## PR Checklist:


sebastiendan:
The failed grpc stub generation workflow run can be discarded as only the path to the files has changed, so we don't need a new npm package version.

crates/topos-api/src/graphql/checkpoint.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+#[derive(Debug, Default, Serialize, Deserialize, InputObject)]
+pub struct SourceCheckpoint {
+    pub source_subnet_ids: Vec<InputSubnetId>,
+    pub positions: Vec<SourceStreamPosition>,
+}
+
+impl InputSubnetId {
+    pub fn to_uci_subnet_id(self) -> topos_uci::SubnetId {
```
hadjiszs, line 30: nit: implementing `into`? or is it to reflect a potential costly conversion?

hadjiszs (approved):

crates/topos-api/src/graphql/checkpoint.rs
gruberb: The problem is that `topos_uci::SubnetId` is in a different module. We would have to do the `From<SubnetId` in the module it is defined, which creates a cyclic dependency. It's not the nicest solution, but I didn't want to do such a big refactor to restructure the modules.

crates/topos-api/src/graphql/checkpoint.rs
```diff
@@ -0,0 +1,43 @@
...
+        // Remove the leading "0x" from the hexadecimal string
+        let hex_string = self.value.trim_start_matches("0x");
+
+        // Decode the hexadecimal string into a vector of bytes
+        let bytes = hex::decode(hex_string).expect("Failed to decode hexadecimal string");
```
atanmarko, line 35: I believe we should not use expect in production code. If it could fail, function should return error.

crates/topos-tce-api/src/graphql/query.rs
```diff
@@ -0,0 +1,65 @@
...
+                            .to_uci_subnet_id(),
+                        position: topos_tce_storage::Position(
+                            from_source_checkpoint.positions[index].position,
+                        ),
+                    },
+                    limit: first,
+                })
+                .await
+                .map_err(|_| GraphQLServerError::StorageError)?;
+            tracing::warn!("Returned from storage: {certificates_with_position:?}");
```
atanmarko, line 42: No need for warning here, debug is probably enough

crates/topos/src/components/tce/mod.rs
```diff
@@ -170,14 +171,18 @@ pub(crate) async fn handle_command(
...
     }
 
-    info!("API gRPC endpoint reachable at {}", config.api_addr);
-    info!("Broadcast Parameters {:?}", config.tce_params);
+    tracing::warn!("API gRPC endpoint reachable at {}", config.api_addr);
+    info!(
+        "API GraphQL endpoint reachable at {}",
+        config.graphql_api_addr
+    );
+    tracing::warn!("Broadcast Parameters {:?}", config.tce_params);
```
atanmarko, line 185: Import `tracing::warn` and use just `warn!()`, then it would be easy to replace some other tracing/log library if needed

crates/topos-api/src/graphql/checkpoint.rs
Freyskeyd: @gruberb It is possible as we already do that for the `gRPC` ones. I will create a PR to fix that :)

crates/topos-api/src/graphql/certificate.rs
```diff
@@ -0,0 +1,49 @@
+use async_graphql::SimpleObject;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize, SimpleObject)]
+pub struct SubnetId {
```
Freyskeyd, line 5: Should have been in a `subnet` file not in `certificate`

crates/topos-api/src/graphql/checkpoint.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::graphql::errors::GraphQLServerError;
+use topos_uci::{self, SUBNET_ID_LENGTH};
+
+#[derive(Debug, Default, Serialize, Deserialize, InputObject)]
+pub struct CertificateId {
+    pub value: String,
+}
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize, InputObject)]
+pub struct InputSubnetId {
```
Freyskeyd, line 13: `InputSubnetId` is the same as `SubnetId` in `certificate` file. We can merge both and have a unique `SubnetId` for input and output

gruberb: We would need to implement `InputObject` and `SimpleObject` by hand then. I thought I leave it like that, since we might have more types in the future like this, so we can have `input` and `output` folders (or sth along these lines)

crates/topos-api/src/graphql/certificate.rs
gruberb: Yeah, the structuring of the folder was a bit rushed. Happy to move things around. I was hoping we get another endpoint or two in soon, so we can figure things out then ones we have more sample data.

crates/topos-api/src/graphql/checkpoint.rs
gruberb: You mean importing the `topos_uci::SubnetId` somewhere else? I didn't want to touch the `gRPC` implementation for a `GraphQL` use case. Otherwise, to `impl` something on a type, AFAIK, you cannot import it! But curious what you mean exactly :)

PR #245, opened 2023-06-16
feat: create config file for subcommands + take values from both config file and cli

gruberb:
# Description

Introducing a `config.toml` file, which can be set with either `default` values or ones passed via CLI arguments. This PR adds two new command line arguments:

* `topos node init`
* `topos node up`

The `init` creates the initial `config.toml` file in the `$TOPOS_HOME` (default to `~/.config/topos`) directory. It will create a folder for each node, `default` being the default one.

```
.config/topos/
  |_default
     |_config.toml
```

Fixes [TP-528](https://linear.app/toposware/issue/TP-528)

## Additions and Changes

* `topos node init`
* `topos node up`

## PR Checklist:


crates/topos/src/config/sequencer.rs
```diff
@@ -0,0 +1,90 @@
...
+    // Core contract address
+    #[serde(default = "default_subnet_contract_address")]
+    pub subnet_contract_address: String,
+
+    /// Base Uri of TCE node to call grpc service api
+    #[serde(default = "default_base_tce_api_url")]
+    pub base_tce_api_url: String,
+
+    /// Polygon subnet node data dir, containing `consensus/validator.key`, e.g. `../test-chain-1`
+    #[serde(default = "default_subnet_data_dir")]
```
gruberb, line 32: Default to this test directory now. We will have to change that and make it required once we add the actual `topos node up` logic to start `tce` and `sequencer`.

crates/topos/src/options.rs
```diff
@@ -29,11 +34,29 @@ pub(crate) struct Opt {
...
+    pub(crate) home: PathBuf,
+
     #[command(subcommand)]
     pub(crate) commands: ToposCommand,
 }
 
-#[derive(Subcommand, Debug)]
+/// If no path is given for the --home argument, we use the default one
+/// ~/.config/topos for a UNIX subsystem
+fn get_default_home() -> OsString {
```
gruberb, line 52: This is currently generating problems with the snapshot tests. Locally, the `OsString` will add the `USERNAME` to the path, whereas in CI, this will result in something different.

Thinking about how we can adjust the tests for this...

Freyskeyd: You can use `regex` to parse the result output in the test and replace the path with a static one such as `home`

crates/topos/src/components/node/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+    }: NodeCommand,
+) -> Result<(), Box<dyn std::error::Error>> {
+    match subcommands {
+        Some(NodeCommands::Init(cmd)) => {
+            let name = cmd.name.expect("No name or default was given");
+
+            // Construct path to node config
+            // will be $TOPOS_HOME/node/default/ with no given name
+            // and $TOPOS_HOME/node/<name>/ with a given name
+            let node_path = home.join("node").join(name);
```
Freyskeyd, line 45: At first I think we don't need to generate the `$TOPOS_HOME/node/default/config.toml` but the `$TOPOS_HOME/node/base.toml` right @hadjiszs ?

crates/topos/src/components/node/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+            println!(
+                "Created node config file at {}/config.toml",
+                node_path.display()
+            );
+
+            Ok(())
+        }
+        Some(NodeCommands::Up(cmd)) => {
+            let name = cmd.name.expect("No name or default was given for node");
+            let node_path = home.join("node").join(name.clone());
```
Freyskeyd, line 89: I think you can pass a ref here to avoid clone

crates/topos/src/config/sequencer.rs
```diff
@@ -0,0 +1,89 @@
...
+use figment::{
+    providers::{Format, Serialized, Toml},
+    Figment,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct SequencerConfig {
+    /// SubnetId of the local subnet node, hex encoded 32 bytes starting with 0x
+    pub subnet_id: Option<String>,
```
Freyskeyd, line 14: I don't think we need all those values as we will refactor and sanitize the CLI by taking only important params. Same for the TCE

crates/topos/src/main.rs
```diff
@@ -16,17 +21,19 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let args = options::Opt::parse();
 
-    match args.commands {
+    match args.commands.clone() {
```
Freyskeyd, line 24: you don't need to clone here

crates/topos/src/options.rs
```diff
@@ -29,11 +34,29 @@ pub(crate) struct Opt {
     )]
     pub(crate) verbose: u8,
 
+    /// Home directory for the configuration
+    #[arg(
+    long,
+    env = "TOPOS_HOME",
+    default_value = get_default_home(),
+    global = true
+    )]
```
Freyskeyd, line 43:
```suggestion
    #[arg(
        long,
        env = "TOPOS_HOME",
        default_value = get_default_home(),
        global = true
    )]
```

crates/topos/src/options.rs
```diff
@@ -29,11 +34,29 @@ pub(crate) struct Opt {
...
     #[command(subcommand)]
     pub(crate) commands: ToposCommand,
 }
 
-#[derive(Subcommand, Debug)]
+/// If no path is given for the --home argument, we use the default one
+/// ~/.config/topos for a UNIX subsystem
+fn get_default_home() -> OsString {
+    let mut home = dirs::home_dir().unwrap();
+    home.push(".config");
```
Freyskeyd, line 54: if we aim to use config folder, we can use `config_dir`  instead of `home_dir`

crates/topos/src/options.rs
```diff
@@ -15,11 +17,14 @@ use crate::components::setup::commands::SetupCommand;
 #[cfg(feature = "subnet")]
 use crate::components::subnet::commands::SubnetCommand;
 
+#[cfg(all(feature = "sequencer", feature = "tce"))]
+use crate::components::node::commands::NodeCommand;
+
 pub(crate) mod input_format;
 
-#[derive(Parser, Debug)]
+#[derive(Parser, Debug, Serialize)]
```
Freyskeyd, line 25:
```suggestion
#[derive(Parser, Debug)]
```

crates/topos/src/options/input_format.rs
```diff
@@ -1,6 +1,7 @@
 use clap::ValueEnum;
+use serde::Serialize;
 
-#[derive(ValueEnum, Copy, Clone, Debug)]
+#[derive(ValueEnum, Copy, Clone, Debug, Serialize)]
```
Freyskeyd, line 4:
```suggestion
#[derive(ValueEnum, Copy, Clone, Debug)]
```

crates/topos/tests/network.rs
```diff
@@ -10,7 +12,13 @@ fn network_spam_help_display() -> Result<(), Box<dyn std::error::Error>> {
 
     let result: &str = std::str::from_utf8(&output.get_output().stdout)?;
 
-    insta::assert_snapshot!(result);
+    // Sanitize the result here:
+    // When run locally, we get /Users/<username>/.config/topos
+    // When testing on the CI, we get /home/runner/.config/topos
+    let pattern = Regex::new(r"\[default: .+?/.config/topos\]").unwrap();
+    let sanitized_result = pattern.replace(result, "[default: /home/runner/.config/topos]");
```
Freyskeyd, line 19: maybe you can put this in a single function

rustfmt.toml
```diff
@@ -1,3 +1,3 @@
 edition = "2021"
 use_field_init_shorthand = true
-reorder_imports = true
+reorder_imports = true
```
Freyskeyd, line 3: need a break line here

Freyskeyd (changes requested):

crates/topos/src/options.rs
gruberb: This gives us a different set of problems with sanitizing for the snapshot tests, because every OS has a different `config` path. I don't see the advantage of adding more sanitizing tests if we can just more tightly control where the default path is.

crates/topos/src/config/sequencer.rs
gruberb: I know, but I also don't exactly know yet which we need and which we don't. So I don't see the value in slimming it down now, when we have to go over it again anyway to decide what goes where!

Freyskeyd: As mentioned before, we don't need anything in the config files for now. It is part of the next iteration to add and architecture that. I would prefer to have a blank page to avoid miss leading pre existing argument.

But I let @hadjiszs decide on that.

crates/topos/src/components/node/mod.rs
gruberb: The read of a specific node already works. But I don't build the `base.toml`, just the specific `<node_name>/config.toml`. I mark this therefore as resolved.

crates/topos/src/components/network/commands.rs
```diff
@@ -1,4 +1,5 @@
 use clap::{Args, Subcommand};
+use serde::Serialize;
```
Freyskeyd, line 2:
```suggestion
```

crates/topos/src/components/network/mod.rs
```diff
@@ -8,6 +9,7 @@ use topos_certificate_spammer::CertificateSpammerConfig;
 use tracing::{error, info};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
+use crate::config::Config;
```
Freyskeyd, line 12:
```suggestion
```

crates/topos/src/components/node/commands/init.rs
```diff
@@ -0,0 +1,22 @@
...
+    /// Role of your node
+    #[arg(long, env = "TOPOS_NODE_ROLE", default_value = "validator")]
+    pub role: Option<String>,
+
+    /// Subnet of your node
+    #[arg(long, env = "TOPOS_NODE_SUBNET", default_value = "topos")]
+    pub subnet: Option<String>,
+}
+
+impl Init {}
```
Freyskeyd, line 22:
```suggestion
```

crates/topos/src/components/node/commands/up.rs
```diff
@@ -0,0 +1,14 @@
...
+
+#[derive(Args, Debug, Serialize)]
+#[command(about = "Spawn your node!")]
+pub struct Up {
+    /// Name to identify your node
+    #[arg(long, env = "TOPOS_NODE_NAME", default_value = "default")]
+    pub name: Option<String>,
+}
+
+impl Up {}
```
Freyskeyd, line 14:
```suggestion
```

crates/topos/src/components/node/commands/up.rs
```diff
@@ -0,0 +1,14 @@
...
+
+use clap::Args;
+use serde::Serialize;
+
+#[derive(Args, Debug, Serialize)]
+#[command(about = "Spawn your node!")]
+pub struct Up {
+    /// Name to identify your node
+    #[arg(long, env = "TOPOS_NODE_NAME", default_value = "default")]
+    pub name: Option<String>,
```
Freyskeyd, line 11: If the `name` is present in every `node` command maybe it should be on the `node` root command. but we could update that later

crates/topos/src/components/tce/mod.rs
Freyskeyd, line 1: I don't think we need all those imports. We may want to add `--all-features` on the CI to check it

crates/topos/src/config/base.rs
```diff
@@ -0,0 +1,59 @@
+use std::path::Path;
+
+use figment::{
+    providers::{Format, Serialized, Toml},
+    Figment,
+};
+use serde::{Deserialize, Serialize};
+
+#[cfg(feature = "node")]
```
Freyskeyd, line 9:
```suggestion
```
you don't need it as the whole flag is under feature in the `mod.rs`

crates/topos/src/options.rs
```diff
@@ -15,11 +17,14 @@ use crate::components::setup::commands::SetupCommand;
...
 
+#[cfg(feature = "node")]
+use crate::components::node::commands::NodeCommand;
+
 pub(crate) mod input_format;
 
 #[derive(Parser, Debug)]
 #[clap(name = "topos", about = "Topos CLI")]
-pub(crate) struct Opt {
+pub struct Opt {
```
Freyskeyd, line 27: Why do we need this to be `pub` instead of `pub(crate)` ?

Freyskeyd (approved):

crates/topos/src/components/node/commands/up.rs
gruberb: Will leave it for next PR!

crates/topos/src/components/tce/mod.rs
gruberb: Removed the not needed imports!

PR #246, opened 2023-06-19
refactor: update graphql/grpc files and compilation

Freyskeyd:
# Description

### Bug fix (non-breaking change which fixes an issue)

- [x] Fix gRPC build compilation

## PR Checklist:


gruberb (approved):

.github/workflows/grpc-client-stubs.yml
```diff
@@ -25,19 +25,19 @@ jobs:
...
+          cache-dependency-path: ./crates/topos-api/proto/package-lock.json
           registry-url: https://registry.npmjs.org
           scope: "@topos-network"
 
       - name: Install grpc-tool globally
         run: npm install -g grpc-tools
 
       - name: Install npm packages
-        working-directory: ./crates/topos-api/src/grpc/proto
+        working-directory: ./crates/topos-api/proto
```
gruberb, line 36: Technically I would still like to move `proto` into `gprc`, since we don't use it anywhere else. If it's one way above, one might assume it is used somehwere else (like `graphql`).

atanmarko (approved):

.github/workflows/grpc-client-stubs.yml
Freyskeyd: @gruberb I can see the point of having the `proto` into the `grpc` folder but for me as it only contains `protobuf` files and are also used by `npm` I would prefer to not have them in the `src` folder which should only contains `rust` related files and files that are use after running the `build.rs`. And also, if someone want to check the protofiles exposed by the `topos-api` having them at the root is easier than digging the code base. Same for Graphql, if we were using a predefined schema, I would put it into another folder at the root of the crate.

gruberb: Thanks for explaining it! Yeah, can agree on that :)

PR #256, opened 2023-06-30
chore: move prometheus behind builder pattern + add env variable

gruberb:
# Description

Cleaning up and moving the prometheus server behind a builder pattern we use with other processes as well.

## Additions and Changes

Added the `TCE_METRICS_API_ADDR` environment variable (CLI + ENV), which defaults to `0.0.0.0:3000`.

## PR Checklist:


crates/topos-metrics/src/server.rs
```diff
@@ -0,0 +1,41 @@
...
+            get(|| async move {
+                let mut buffer = Vec::new();
+                let encoder = TextEncoder::new();
+
+                // Gather the metrics.
+                let metric_families = prometheus::gather();
+                // Encode them to send.
+                encoder.encode(&metric_families, &mut buffer).unwrap();
+
+                String::from_utf8(buffer.clone()).unwrap()
```
Freyskeyd, line 33: Maybe we can just put this closure into a `pub` function of the `metrics` crate and define the server in `topos-tce-api` as it is the place where we have every api server definnition

gruberb: Hm, yeah... Then we split it in three different pieces: `topos-metrics`, `topos-tce-api`, and in `topos-tce`, we built and start the server. Does it then make sense to start the server also in `topos-tce-api`?

Freyskeyd: `topos-metrics` is a little crate to handle metrics system, nothing more.
`topos-api` exposes external api structs.
`topos-tce` is the runtime of a TCE node.

For now, the `API` on port `3000` is just here to expose a `/metrics` which read the `topos-metrics` state. But at some point we could expose other things on that, like a `/status` or any other non-related metrics things. It's a matter of SOC.

gruberb: 👍 I updated the PR!

crates/topos-metrics/src/lib.rs
```diff
@@ -31,3 +31,15 @@ lazy_static! {
     pub static ref CERTIFICATE_DELIVERED: IntCounter =
         register_int_counter!("certificate_delivered", "Number of certificate delivered.").unwrap();
 }
+
+pub async fn gather_metrics() -> String {
```
Freyskeyd, line 35: I don't see why this is `async`, is it to directly use it as a `handler` in `axum` to avoid wrapping a sync method in an `async { .. }` ?

Freyskeyd (approved):

crates/topos-metrics/src/lib.rs
gruberb: Yes! But I made it `sync` now and moved it in an `async` block 👍  Much more obvious this way, thanks!

PR #260, opened 2023-07-11
feat: adding better port detection

Freyskeyd:
# Description

Improving port detection and reservation for tests.

## PR Checklist:

### DIFF
--- a/crates/topos-test-sdk/src/networking/mod.rs
+++ b/crates/topos-test-sdk/src/networking/mod.rs
@@ -1,0 +1,22 @@
+use std::net::SocketAddr;
+use std::net::{TcpListener, TcpStream};
+
+pub fn get_available_port() -> u16 {
+    get_available_addr().port()
+}
+
+pub fn get_available_addr() -> SocketAddr {
+    let host = "127.0.0.1";
+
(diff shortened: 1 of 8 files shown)


crates/topos-test-sdk/src/networking/mod.rs
```diff
@@ -0,0 +1,22 @@
+use std::net::SocketAddr;
+use std::net::{TcpListener, TcpStream};
+
+pub fn get_available_port() -> u16 {
+    get_available_addr().port()
+}
+
+pub fn get_available_addr() -> SocketAddr {
+    let host = "127.0.0.1";
```
gruberb, line 9: Is this changed to `127.0.0.1` on purpose?

gruberb (commented):
Nice find! Just wondering why the move to `127.0.0.1` instead of `0.0.0.0`.

crates/topos-test-sdk/src/networking/mod.rs
Freyskeyd: Yes I want to provide `addr` only for the loopback interface

gruberb (approved):

PR #268, opened 2023-07-27
fix: prune the buffers after transfer

hadjiszs:
# Description

Fix by pruning the applied buffers once transferred to the related task

## PR Checklist:


gruberb:
Instead of looping twice through all messages and tasks, we can add this in the `match Broadcast` arm in each `TaskManager`, after `task.run()` or `task.into_future()`:

```rust
if let Some(messages) = self.buffered_messages.remove(&cert.id) {
    for msg in messages {
         _ = task_context.sink.send(msg).await;
     }
}
```

hadjiszs:
closing in favor of #269

PR #282, opened 2023-08-22
feat: removing the config folder and exit early if edge-config fails

gruberb:
# Description

The `polygon-edge` binary is being run on `topos node init`. This can fail for various reason. If it does, we don't want to write a config file, we want to fail and tell the user that it failed. 

Fixes: TP-662

## PR Checklist:


atanmarko:
@gruberb  Update the PR description?

crates/topos/src/components/node/mod.rs
```diff
@@ -79,7 +79,13 @@ pub(crate) async fn handle_command(
             let mut config_toml = toml::Table::new();
 
             // Generate the Edge configuration
-            let handle = services::generate_edge_config(edge_path, node_path.clone());
+            if let Ok(result) = generate_edge_config(edge_path, node_path.clone()).await {
+                if result.is_err() {
+                    println!("Failed to generate edge config");
+                    remove_dir_all(home).expect("failed to remove config folder");
```
atanmarko, line 85: Are we sure we want to remove whole topos config folder if generation config fails?

gruberb: I should have used `Draft` for this PR, sorry. Yeah it's not the final one yet, just a first working solution. Have to go through the cases what we want to remove again, or if we have to remove at all.

hadjiszs (approved):

PR #288, opened 2023-08-29
feat: add clock (time)

Freyskeyd:
# Description

This PR initiates the work on the state sync by introducing a `clock` in order to follow the block production. For now we use a time based clock.

## PR Checklist:

### DIFF
--- a/crates/topos-clock/Cargo.toml
+++ b/crates/topos-clock/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "topos-clock"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+tokio.workspace = true
+futures.workspace = true
+thiserror.workspace = true
+chrono = {version = "0.4", default-features = false, features = ["clock"]}
+tracing.workspace = true
--- a/crates/topos-clock/src/lib.rs
+++ b/crates/topos-clock/src/lib.rs
@@ -1,0 +1,58 @@
+//! This crate is responsible for managing the clock pace.
+//!
+//! The Clock is responsible of giving informations about Epoch and Delta timing by exposing
+//! reference to the data but also by broadcasting `EpochChange` events.
+
+use std::sync::{atomic::AtomicU64, Arc};
+
+use tokio::sync::broadcast;
+
+mod time;
+
+pub use time::TimeClock;
+
+const BROADCAST_CHANNEL_SIZE: usize = 100;
+
+pub trait Clock {
+    /// Compute Epoch/Block numbers and spawn the clock task.
+    fn spawn(self) -> Result<broadcast::Receiver<Event>, Error>;
+    /// Return a reference to the current block number
+    fn block_ref(&self) -> Arc<AtomicU64>;
+    /// Return a reference to the current epoch number
+    fn epoch_ref(&self) -> Arc<AtomicU64>;
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Event {
+    EpochChange,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Unable to generate spawn date")]
+    SpawnDateFailure,
+}
+
+#[cfg(test)]
+mod tests {
+    use chrono::{Duration, Utc};
+
+    use crate::{Clock, Event, TimeClock};
+
+    #[tokio::test]
+    async fn test_time_clock() {
+        let genesis = Utc::now()
+            .checked_sub_signed(Duration::seconds(30))
+            .unwrap();
+
+        let clock = TimeClock::new(genesis, 5).unwrap();
+        let current_block = clock.block_ref();
+        let current_epoch = clock.epoch_ref();
+
+        let mut recv = clock.spawn().unwrap();
+
+        assert_eq!(recv.recv().await, Ok(Event::EpochChange));
+        assert_eq!(current_epoch.load(std::sync::atomic::Ordering::Relaxed), 7);
+        assert_eq!(current_block.load(std::sync::atomic::Ordering::Relaxed), 30);
+    }
+}
--- a/crates/topos-clock/src/time.rs
+++ b/crates/topos-clock/src/time.rs
@@ -1,0 +1,104 @@
+use std::{
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+use chrono::{DateTime, Utc};
+use tokio::{
+    spawn,
+    sync::broadcast,
+    time::{interval_at, Instant},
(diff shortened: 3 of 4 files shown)


gruberb (approved):

crates/topos-clock/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "topos-clock"
+version = "0.1.0"
+edition = "2021"
+
```
dvdplm, line 5: I for one would appreciate a `description` entry here, along with license and author fields.

crates/topos-clock/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
dvdplm, line 6:
```suggestion

```

crates/topos-clock/src/lib.rs
```diff
@@ -0,0 +1,58 @@
+//! This crate is responsible for managing the clock pace
```
dvdplm, line 1: The PR description is a bit terse, but it sounds like this might be a temporary solution. I still think we should have a more complete top-level justification for what this crate is doing. Especially important imo is to indicate to readers if this code is a stop-gap solution until we have something better in place (if that is indeed the case).

crates/topos-clock/src/time.rs
```diff
@@ -0,0 +1,99 @@
...
+use tokio::{
+    spawn,
+    sync::broadcast,
+    time::{interval_at, Instant},
+};
+use tracing::info;
+
+use crate::{Clock, Error, Event};
+
+pub struct TimeClock {
```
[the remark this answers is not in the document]
dvdplm: Agreed. Question about the name: are we expecting further implementations of the `Clock` trait, so that "TimeClock" is but one implementation of it? Can you elaborate a bit on what other `Clock` implementations we anticipate?

crates/topos-clock/src/time.rs
```diff
@@ -0,0 +1,99 @@
...
+    genesis: DateTime<Utc>,
+    current_block: Arc<AtomicU64>,
+    epoch_duration: u64,
+    current_epoch: Arc<AtomicU64>,
+}
+
+impl Clock for TimeClock {
+    fn spawn(mut self) -> Result<broadcast::Receiver<Event>, Error> {
+        let (sender, receiver) = broadcast::channel(100);
```
dvdplm, line 28: Is the capacity set to `100` for some specific reason or how was this value chosen? Should it be a constant perhaps?

crates/topos-clock/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use std::sync::{atomic::AtomicU64, Arc};
+
+use tokio::sync::broadcast;
+
+mod time;
+
+pub use time::TimeClock;
+
+pub struct ClockManager {}
```
dvdplm, line 11: missing docs

crates/topos-clock/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+impl ClockManager {
+    pub fn spawn<C: Clock>(clock: C) -> Result<broadcast::Receiver<Event>, Error> {
+        clock.spawn()
+    }
+}
+
+pub trait Clock {
+    fn spawn(self) -> Result<broadcast::Receiver<Event>, Error>;
+    fn block_ref(&self) -> Arc<AtomicU64>;
+    fn epoch_ref(&self) -> Arc<AtomicU64>;
```
dvdplm, line 22: Missing docs

crates/topos-clock/Cargo.toml
Freyskeyd: I will create a ticket to standardize all the `Cargo.toml` in order to have those fields

dvdplm: Yes please, that would be lovely. :)

crates/topos-clock/src/time.rs
```diff
@@ -0,0 +1,103 @@
...
+use tokio::{
+    spawn,
+    sync::broadcast,
+    time::{interval_at, Instant},
+};
+use tracing::info;
+
+use crate::{Clock, Error, Event, BROADCAST_CHANNEL_SIZE};
+
+/// Time based clock implementation.
```
dvdplm, line 19:
```suggestion
/// Time based clock implementation, mostly intended for tests.
```

Right?

crates/topos-clock/src/lib.rs
```diff
@@ -0,0 +1,57 @@
+//! This crate is responsible for managing the clock pace.
+//!
+//! The Clock is responsible of giving informations about Epoch and Delta timing by exposing
```
dvdplm, line 3:
```suggestion
//! The Clock is responsible of giving informations about Epoch and Delta timings by exposing
```

Is "delta' the same as "block time"?

crates/topos-clock/src/time.rs
```diff
@@ -0,0 +1,103 @@
...
+    fn block_ref(&self) -> Arc<AtomicU64> {
+        self.current_block.clone()
+    }
+    fn epoch_ref(&self) -> Arc<AtomicU64> {
+        self.current_epoch.clone()
+    }
+}
+
+impl TimeClock {
+    pub fn new(genesis: DateTime<Utc>, epoch_duration: u64) -> Result<Self, Error> {
```
dvdplm, line 53: Missing docs.

crates/topos-clock/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+use tokio::sync::broadcast;
+
+mod time;
+
+pub use time::TimeClock;
+
+const BROADCAST_CHANNEL_SIZE: usize = 100;
+
+pub trait Clock {
+    fn spawn(self) -> Result<broadcast::Receiver<Event>, Error>;
```
dvdplm, line 17: Missing docs.

crates/topos-clock/src/time.rs
Freyskeyd: Not really, until we have blockchain integration it will be the only implementation.

crates/topos-clock/src/lib.rs
Freyskeyd: No, `Delta` are an amount of blocks. The amount of blocks isn't specified for now.

dvdplm: got it, thanks

crates/topos-clock/src/lib.rs
Freyskeyd: @dvdplm fixed

dvdplm (approved):

PR #298, opened 2023-09-08
feat: add validator id and verification to echo and ready messages

gruberb:
# Description

Introducing `ValidatorId` to verify that `ECHO | READY` message come from a valid `validator`. We also add a signature to `EACH | READY` messages. Therefore the `double_echo` has to be aware of the `validator` private key (secret). 

Fixes TP-677 and TP-678

## Additions and Changes

- Adding `signing_key` to the `TceConfig`
- Adding `signature` and `validator_id` to `DoubleEchoCommand` and `ProtocolEvents::Echo` and `ProtocolEvents::Ready`. 

## PR Checklist:


crates/topos-tce-broadcast/src/double_echo/mod.rs
```diff
@@ -25,6 +25,10 @@ pub struct DoubleEcho {
     task_manager_message_sender: mpsc::Sender<DoubleEchoCommand>,
     /// The overview of the network, which holds echo and ready subscriptions and the network size
     pub subscriptions: SubscriptionsView,
+    /// Public ETH address
+    pub authority_id: AuthorityId,
+    /// Known and approved validators in the network
+    pub validators: Vec<String>,
```
hadjiszs, line 31: I think that you want this set of `AuthorityId` in the sampler directly so that your different HashSet in `subscriptions` are directly fed with `AuthorityId`s no?

gruberb: I am not sure about this part. Since we don't want to touch the `libp2p` part where the discovery is done through `PeerId`s. I have to go through thoroughly and check where exactly we need `PeerId`s and where we use `AuthorityId`.

crates/topos-tce-transport/src/lib.rs
```diff
@@ -31,6 +32,41 @@ impl ReliableBroadcastParams {
...
+    fn from(address: &str) -> Self {
+        // Remove the "0x" prefix from the address
+        let address_without_prefix = &address[2..];
+
+        // Convert the hexadecimal address to bytes
+        let bytes = hex::decode(address_without_prefix).expect("Failed to decode hex string");
+
+        // Ensure the bytes have the correct length (20 bytes)
+        if bytes.len() != 20 {
+            panic!("Invalid address length");
```
gruberb, line 56: Have to return a proper `Error` still.

crates/topos-tce-broadcast/src/lib.rs
```diff
@@ -82,13 +82,17 @@ pub enum DoubleEchoCommand {
...
+        signature: Signature,
     },
 
     /// When ready reply received
     Ready {
         from_peer: PeerId,
+        validator_id: ValidatorId,
         certificate_id: CertificateId,
+        signature: Signature,
     },
```
dvdplm, line 96: Same question as elsewhere: do you think we should have types for this data so we never accidentally build an `Echo` with data from a `Ready`?

crates/topos-tce-broadcast/src/task_manager_channels/mod.rs
```diff
@@ -1,7 +1,7 @@
+use ethers::prelude::LocalWallet;
```
dvdplm, line 1: Why not import this from `ethers-signers`?

crates/topos-tce-broadcast/src/double_echo/mod.rs
```diff
@@ -25,6 +27,12 @@ pub struct DoubleEcho {
     task_manager_message_sender: mpsc::Sender<DoubleEchoCommand>,
     /// The overview of the network, which holds echo and ready subscriptions and the network size
     pub subscriptions: SubscriptionsView,
+    /// Public ETH address
+    pub validator_id: ValidatorId,
+    /// Keypair to sign and verify ECHO and READY messages
+    pub wallet: LocalWallet,
```
dvdplm, line 33: Random question: the `Wallet` contains the address already; are we checking that it is the same one as the `ValidatorId`? Could we somehow use the wallet address instead of having to use our own field everywhere?

crates/topos-tce-transport/src/lib.rs
```diff
@@ -117,3 +134,62 @@ pub enum ProtocolEvents {
...
+    certificate_id: CertificateId,
+    wallet: LocalWallet,
+) -> Result<Signature, WalletError> {
+    let mut hash = Vec::new();
+    hash.extend(certificate_id.as_array().iter().cloned());
+    hash.extend(validator_id.as_bytes());
+
+    let hash = keccak256(hash);
+
+    wallet.sign_message(hash.as_slice()).await
```
dvdplm, line 180: This is going to hash the message, our `hash`, one more time. See https://docs.rs/ethers-core/2.0.10/src/ethers_core/utils/hash.rs.html#14

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
```diff
@@ -57,28 +63,36 @@ impl BroadcastState {
...
         // any Echo or Ready messages
         // Sending our Echo message
         if let Status::Pending = self.status {
             _ = self.event_sender.try_send(ProtocolEvents::Echo {
                 certificate_id: self.certificate.id,
+                signature: sign_message(
+                    self.validator_id.clone(),
+                    self.certificate.id,
+                    self.wallet.clone(),
+                )
```
dvdplm, line 92: I think `update_status` is a fairly hot function yes? Perhaps it makes sense to look into avoiding these clones and optimize `sign_message` a bit. Perhaps we can sign with `self.wallet` directly?

crates/topos-tce-transport/src/lib.rs
```diff
@@ -117,3 +134,62 @@ pub enum ProtocolEvents {
...
+
+impl From<H160> for ValidatorId {
+    fn from(address: H160) -> Self {
+        ValidatorId(address)
+    }
+}
+
+impl From<&str> for ValidatorId {
+    fn from(address: &str) -> Self {
+        ValidatorId(H160::from_str(address).expect("Cannot transform to H160"))
```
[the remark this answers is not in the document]
gruberb: Fair point. I wasn't sure, since:
* If we can't parse it here, the node doesn't function correctly 
* This `from` is being used just in the CLI, and can be removed once we use the `genesis.json` in the CI.

crates/topos-tce-transport/src/lib.rs
```diff
@@ -117,3 +134,62 @@ pub enum ProtocolEvents {
...
+
+    wallet.sign_message(hash.as_slice()).await
+}
+
+pub fn verify_signature(
+    signature: Signature,
+    validator_id: ValidatorId,
+    certificate_id: CertificateId,
+) -> Result<(), SignatureError> {
+    let mut hash = Vec::new();
```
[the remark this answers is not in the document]
gruberb: I tried to keep the usage of external crates to a minimum. `ether-rs` doesn't export `Keccak` anymore. I assume `tiny-keccak` does. If we save memory allocation, possibly worth it!

crates/topos-tce-broadcast/src/double_echo/mod.rs
gruberb: I think the idea was to be chain agnostic. So we might not always have Ethereum public addresses? But yeah, totally see that reasoning. @hadjiszs might have some thoughts on this as well.

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
```diff
@@ -1,6 +1,8 @@
+use crate::sampler::SubscriptionsView;
+use ethers::signers::LocalWallet;
```
Freyskeyd, line 2: This should be part of `topos-crypto` not directly the `topos-tce-broadcast`

crates/topos-tce-transport/src/lib.rs
```diff
@@ -101,14 +114,80 @@ pub enum ProtocolEvents {
...
+    }
+}
+
+impl std::fmt::Display for ValidatorId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "0x{}", hex::encode(self.0))
+    }
+}
+
+pub async fn sign_message(
```
Freyskeyd, line 167: Should be in `topos-crypto` 

Also, this method doesn't need to be `async`. It should accept a `&[u8]` and only `hash` and `sign`. The payload should be defined outside this method and not be strongly-coupled

crates/topos-tce-transport/src/lib.rs
```diff
@@ -101,14 +114,80 @@ pub enum ProtocolEvents {
...
+    certificate_id: CertificateId,
+    wallet: LocalWallet,
+) -> Result<Signature, WalletError> {
+    let mut preimg = Vec::new();
+    preimg.extend(certificate_id.as_array().iter().cloned());
+    preimg.extend(validator_id.as_bytes());
+
+    let hash = keccak256(preimg);
+
+    wallet.sign_message(hash.as_slice()).await
```
Freyskeyd, line 178: Can we switch to the non async methods with is to [hash](https://docs.rs/ethers/latest/ethers/utils/fn.hash_message.html) and then [sign_hash](https://docs.rs/ethers/latest/ethers/signers/struct.Wallet.html?search=sign_me#method.sign_hash) ?

crates/topos-tce-broadcast/src/double_echo/mod.rs
```diff
@@ -145,8 +162,30 @@ impl DoubleEcho {
...
-                                DoubleEchoCommand::Echo { from_peer, certificate_id } => self.handle_echo(from_peer, certificate_id).await,
-                                DoubleEchoCommand::Ready { from_peer, certificate_id } => self.handle_ready(from_peer, certificate_id).await,
+                                DoubleEchoCommand::Echo { from_peer, certificate_id, validator_id, signature } => {
+                                    // Check if source is part of known_validators
+                                    if !self.validators.contains(&validator_id) {
+                                        return error!("ECHO message comes from non-validator: {}", validator_id);
+                                    }
+
+                                    if let Err(e) = verify_signature(signature, validator_id.clone(), certificate_id) {
+                                        return error!("ECHO messag signature cannot be verified from: {}", e);
```
Freyskeyd, line 172:
```suggestion
                                        return error!("ECHO message signature cannot be verified from: {}", e);
```

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
```diff
@@ -65,7 +71,7 @@ impl BroadcastState {
             });
         }
 
-        state.update_status();
+        state.update_status().await;
```
Freyskeyd, line 74: I would want to keep the `BroadcastState` non async as it is somehow a state machine.

crates/topos-tce-transport/src/lib.rs
```diff
@@ -101,14 +114,80 @@ pub enum ProtocolEvents {
...
 
     /// Stable Sample
     StableSample,
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum Error {}
+
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq, Hash)]
+pub struct ValidatorId(H160);
```
Freyskeyd, line 137: This type is `Copy` you can add the `#[derive(Copy)]`

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
```diff
@@ -92,30 +98,38 @@ impl BroadcastState {
...
-    fn update_status(&mut self) -> Option<Status> {
+    async fn update_status(&mut self) -> Option<Status> {
         // Nothing happened yet, we're in the initial state and didn't Procced
         // any Echo or Ready messages
         // Sending our Echo message
         if let Status::Pending = self.status {
             _ = self.event_sender.try_send(ProtocolEvents::Echo {
                 certificate_id: self.certificate.id,
+                signature: sign_message(
+                    self.validator_id.clone(),
```
Freyskeyd, line 126: You don't need to clone if you put `ValidatorId` as `Copy`

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
```diff
@@ -92,30 +98,38 @@ impl BroadcastState {
...
+                    self.certificate.id,
+                    self.wallet.clone(),
```
Freyskeyd, line 128: Switching to `non-async` would avoid the need to clone

crates/topos-tce-transport/src/lib.rs
```diff
@@ -101,14 +114,80 @@ pub enum ProtocolEvents {
...
+impl std::fmt::Display for ValidatorId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "0x{}", hex::encode(self.0))
+    }
+}
+
+pub async fn sign_message(
+    validator_id: ValidatorId,
+    certificate_id: CertificateId,
+    wallet: LocalWallet,
```
Freyskeyd, line 170: You don't need ownership to sign here
```suggestion
    wallet: &LocalWallet,
```

crates/topos-tce-transport/src/lib.rs
```diff
@@ -101,14 +114,80 @@ pub enum ProtocolEvents {
...
+    preimg.extend(certificate_id.as_array().iter().cloned());
+    preimg.extend(validator_id.as_bytes());
+
+    let hash = keccak256(preimg);
+
+    wallet.sign_message(hash.as_slice()).await
+}
+
+pub fn verify_signature(
+    signature: Signature,
```
Freyskeyd, line 182: same as above, should verify a `&[u8]`

crates/topos-tce-transport/src/lib.rs
```diff
@@ -101,14 +114,80 @@ pub enum ProtocolEvents {
...
+    }
+}
+
+impl From<H160> for ValidatorId {
+    fn from(address: H160) -> Self {
+        ValidatorId(address)
+    }
+}
+
+impl From<&str> for ValidatorId {
```
Freyskeyd, line 155: Switch to `TryFrom` if possible to avoid `expect`

crates/topos-tce-transport/src/lib.rs
```diff
@@ -101,14 +114,80 @@ pub enum ProtocolEvents {
...
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq, Hash)]
+pub struct ValidatorId(H160);
+
+impl ValidatorId {
+    pub fn as_bytes(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+
+    pub fn address(&self) -> Address {
+        Address::from_str(&format!("0x{}", hex::encode(self.0))).unwrap()
```
Freyskeyd, line 145: if `Address` is `ethers::types::Address` then this method is somehow useless as `Address` is an alias to `H160`

crates/topos-tce-broadcast/src/task_manager_futures/mod.rs
```diff
@@ -56,7 +59,9 @@ impl TaskManager {
         task_completion_sender: mpsc::Sender<(CertificateId, TaskStatus)>,
         subscription_view_receiver: mpsc::Receiver<SubscriptionsView>,
         event_sender: mpsc::Sender<ProtocolEvents>,
+        validator_id: ValidatorId,
         thresholds: ReliableBroadcastParams,
+        wallet: LocalWallet,
```
Freyskeyd, line 64: `LocalWallet` will never change and doesn't need to be mutable. Put it in `Arc<LocalWallet>` to avoid clone

crates/topos-tce-broadcast/src/lib.rs
Freyskeyd: @dvdplm The serialization makes it near-impossible, but we'll have to refactor the message system at some point.

crates/topos-tce-broadcast/src/double_echo/mod.rs
```diff
@@ -145,8 +162,30 @@ impl DoubleEcho {
...
+                                    self.handle_echo(from_peer, certificate_id, validator_id, signature).await
+                                },
+                                DoubleEchoCommand::Ready { from_peer, certificate_id, validator_id, signature } => {
+                                    // Check if source is part of known_validators
+                                    if !self.validators.contains(&validator_id) {
+                                        return error!("READY message comes from non-validator: {}", validator_id);
+                                    }
+
+                                    if let Err(e) = verify_signature(signature, validator_id.clone(), certificate_id) {
+                                        return error!("ECHO messag signature cannot be verified from: {}", e);
```
Freyskeyd, line 184:
```suggestion
                                        return error!("READY message signature cannot be verified from: {}", e);
```

crates/topos-test-sdk/src/tce/protocol.rs
```diff
@@ -1,25 +1,38 @@
...
     tce_params: ReliableBroadcastParams,
-    peer_id: String,
     storage: Arc<ValidatorStore>,
     sender: broadcast::Sender<CertificateDeliveredWithPositions>,
 ) -> (
     ReliableBroadcastClient,
     impl Stream<Item = ProtocolEvents> + Unpin,
 ) {
-    let config = ReliableBroadcastConfig { tce_params };
+    let wallet: LocalWallet = "47d361f6becb933a77d7e01dee7b1c1859b656adbd8428bf7bf9519503e5d5d6"
```
Freyskeyd, line 20: can we have that as a `const` ?

crates/topos/src/components/node/services.rs
```diff
@@ -102,7 +104,7 @@ pub(crate) fn spawn_tce_process(
         version: env!("TOPOS_VERSION"),
     };
 
-    debug!("TCE args: {tce_config:?}");
+    error!("TCE args: {tce_config:?}");
```
Freyskeyd, line 107: Why switching to `error` ?

Freyskeyd (changes requested):

crates/topos-tce-broadcast/src/double_echo/mod.rs
dvdplm: It seems like we're assuming that addresses are ethereum style `H160` most places anyway. And we assume EVM compatibility in many places yeah?

crates/topos-tce-broadcast/src/lib.rs
Freyskeyd: Relying on protobuff types and use Envelope to gather metadata and payload. Didn't design this in detail yet.

crates/topos-crypto/src/messages.rs
```diff
@@ -0,0 +1,39 @@
...
+    wallet: &LocalWallet,
+) -> Result<Signature, WalletError> {
+    let mut message: Vec<u8> = Vec::with_capacity(certificate_id.len() + validator_id.len());
+
+    message.extend_from_slice(certificate_id);
+    message.extend_from_slice(validator_id);
+
+    let hash = hash_message(message.as_slice());
+
+    println!("hash: {:#?}", hash);
```
dvdplm, line 17: Forgot to remove?

crates/topos-crypto/src/messages.rs
```diff
@@ -0,0 +1,39 @@
...
+    public_key: Address,
+) -> Result<(), SignatureError> {
+    let mut message: Vec<u8> = Vec::with_capacity(certificate_id.len() + validator_id.len());
+
+    message.extend_from_slice(certificate_id);
+    message.extend_from_slice(validator_id);
+
+    let hash = hash_message(message.as_slice());
+
+    println!("hash: {:#?}", hash);
```
dvdplm, line 35: same

crates/topos-tce-transport/src/lib.rs
gruberb: It's just a convience function to not to have to to `validator_id.0`.

crates/topos-tce-transport/src/lib.rs
gruberb: In this case, we want to abort however. This function is used when parsing the validatos from `genesis.json` or from the `CLI`. So in case we can't parse the provided data, we should exit right away.

crates/topos-tce-broadcast/src/double_echo/mod.rs
gruberb: Yeah we do. The idea is that at some point we don't need to rely on Ethereum and can plug in non-Blockchain networks as well. But yeah, design decision. I  can see pros and cons on both sides.

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
gruberb: Fair point. But in this case, we just import the `type`. We can `alias` it in `topos-crypto`, but this is just one more indirection in the codebase! Or what do you mean, creating our own version of `LocalWallet`, instantiate it during `tce` and living inside `topos-crypto`?

crates/topos-tce-broadcast/src/double_echo/mod.rs
gruberb: I open an issue so we can discuss this in another PR?

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
gruberb: Solved it by a new type, `MessageSigner`.

crates/topos-crypto/src/messages.rs
```diff
@@ -0,0 +1,55 @@
...
+pub struct MessageSigner {
+    pub public_address: Address,
+    wallet: LocalWallet,
+}
+
+impl MessageSigner {
+    pub fn new(private_key: &str) -> Arc<Self> {
+        let wallet: LocalWallet = private_key
+            .parse()
+            .expect("Cannot create wallet for private key");
```
Freyskeyd, line 17: Maybe replace the `expect` by a proper `MessageSignerError`

```rust
enum MessageSignerError {
	#[error("Unable to parse private key")
	PrivateKeyParsing
}
```

crates/topos-crypto/src/messages.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    pub fn sign_message(
+        &self,
+        certificate_id: &[u8],
+        validator_id: &[u8],
+    ) -> Result<Signature, WalletError> {
+        let mut message: Vec<u8> = Vec::with_capacity(certificate_id.len() + validator_id.len());
+        message.extend_from_slice(certificate_id);
+        message.extend_from_slice(validator_id);
```
Freyskeyd, line 31: As discuss previously, I would prefer to have a method that sign a payload directly, not matter what's inside.
Having that let us iterate over how we compress or encode our payload without the need of updating this method as well.

At some point, the payload itself will need to be serialized before signing it. It's not up to this method to construct the payload.

This method could just accept any type `T` that implement `Into<Vec<u8>>` at the end.

crates/topos-tce-broadcast/Cargo.toml
```diff
@@ -11,27 +11,28 @@ byteorder.workspace = true
 futures.workspace = true
 lazy_static.workspace = true
 rand.workspace = true
+ethers.workspace = true
```
Freyskeyd, line 14: we still need `ethers` here?

crates/topos-tce-broadcast/src/double_echo/mod.rs
```diff
@@ -28,6 +29,12 @@ pub struct DoubleEcho {
     task_manager_message_sender: mpsc::Sender<DoubleEchoCommand>,
     /// The overview of the network, which holds echo and ready subscriptions and the network size
     pub subscriptions: SubscriptionsView,
+    /// Public ETH address
```
Freyskeyd, line 32:
```suggestion
    /// Local node ValidatorId
```

crates/topos-tce-broadcast/src/double_echo/mod.rs
Freyskeyd: I still see the message `e` ^^

crates/topos-tce-broadcast/src/lib.rs
```diff
@@ -3,22 +3,20 @@
 //! Abstracted from actual transport implementation.
 //! Abstracted from actual storage implementation.
 
+use double_echo::DoubleEcho;
+use ethers::types::Signature;
```
Freyskeyd, line 7: re-export the `Signature` from `topos-crypto` to avoid scope deps to `topos-crypto` only

crates/topos-tce-transport/src/lib.rs
```diff
@@ -101,14 +112,61 @@ pub enum ProtocolEvents {
...
+#[derive(Debug, Error)]
+pub enum ValidatorIdConversionError {
+    #[error("Failed to parse address string as H160")]
+    ParseError,
+    #[error("Failed to convert byte array into H160")]
+    InvalidByteLength,
+}
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, PartialEq, Hash)]
+pub struct ValidatorId(H160);
```
Freyskeyd, line 140: I'm wondering if the `ValidatorId` should be part of `tce-transport` or `core`. Let's see that in a next PR

tools/docker-compose.yml
Freyskeyd, line 1: You can revert all those changes

crates/topos/tests/cert_delivery.rs
Freyskeyd, line 1: Is switching to `7` sec do something on the test or was it just to debug?

Freyskeyd (changes requested):

crates/topos-crypto/src/messages.rs
gruberb: Then the calling method has to construct the payload? What would be the signature like in your mind?

I see it as a first step. Once we need a more general signing message method, we do it?

Freyskeyd (approved):

dvdplm:
🎉

PR #299, opened 2023-09-11
fix: fixing gossip registry metric setup

Freyskeyd:
# Description

This PR is fixing a regression introduced during the benchmarks phase leading to some test failing due to concurrency on the `Gossipsub` metrics system.

The option was to make the `build` method of `gossip` behavior `async` or to optionally disable metrics on gossip.

The current option is to move to `async`, if we add or need to test metrics system we will have to rework this part.

## PR Checklist:

- [x]  I have made corresponding changes to the documentation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,9 @@ members = [
     "crates/*"
 ]
 
+[profile.release]
+strip = true
+
 [workspace.dependencies]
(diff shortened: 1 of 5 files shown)


dvdplm (approved):
LGTM.

gruberb (approved):
👍

atanmarko (approved):

PR #305, opened 2023-09-14
feat: fix genesis block sync and add related tests

atanmarko:
# Description

Fixed subnet sync. It now starts from genesis block 0.
New sequencer regressing test checking sync from genesis block and particular historical block.

Fixes TP-706

## PR Checklist:

### DIFF
--- a/crates/topos-sequencer-subnet-runtime/src/proxy.rs
+++ b/crates/topos-sequencer-subnet-runtime/src/proxy.rs
@@ -103,7 +103,7 @@ impl SubnetRuntimeProxy {
             let runtime_proxy = runtime_proxy.clone();
             let subnet_contract_address = subnet_contract_address.clone();
             tokio::spawn(async move {
-                let mut latest_acquired_subnet_block_number = 0;
+                let mut latest_acquired_subnet_block_number: i128 = -1;
 
                 {
                     // To start producing certificates, we need to know latest delivered or pending certificate id from TCE
@@ -118,10 +118,11 @@ impl SubnetRuntimeProxy {
                                     "Source head certificate id received {:?}",
                                     certificate_and_position
                                 );
+                                // If the position is not provided, it should start form -1, so that first fetched is subnet genesis block
                                 let cert_id = certificate_and_position.map(|(id, _position)| id);
-                                let position = certificate_and_position
-                                    .map(|(_id, position)| position)
-                                    .unwrap_or_default();
+                                let position: i128 = certificate_and_position
+                                    .map(|(_id, position)| position as i128)
+                                    .unwrap_or(-1);
                                 // Certificate generation is now ready to run
                                 certification.last_certificate_id = cert_id;
                                 latest_acquired_subnet_block_number = position;
@@ -152,12 +153,12 @@ impl SubnetRuntimeProxy {
 
                 // Sync missing blocks
                 loop {
-                    let current_subnet_block_number: Option<u64> = loop {
+                    let current_subnet_block_number: Option<i128> = loop {
                         tokio::select! {
                             block_number = subnet_listener.get_subnet_block_number() => {
                                 match block_number {
                                     Ok(block_number) => {
-                                        break Some(block_number);
+                                        break Some(block_number as i128);
                                     }
                                     Err(e) => {
                                         error!("Failed to get subnet block number: {:?}", e);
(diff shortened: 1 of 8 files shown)


crates/topos-sequencer-subnet-runtime/tests/subnet_contract.rs
```diff
@@ -921,3 +921,98 @@ async fn test_subnet_send_token_processing(
...
+        panic!("Expected event not received");
+    };
+
+    // Set big timeout to prevent flaky fails. Instead fail/panic early in the test to indicate actual error
+    if tokio::time::timeout(std::time::Duration::from_secs(60), assertion)
+        .await
+        .is_err()
+    {
+        panic!("Timeout waiting for command");
+    }
```
hadjiszs, line 1013: Isn't it redundant with the `#[timeout(std::time::Duration::from_secs(60))]` at the test level?

crates/topos-sequencer-subnet-runtime/tests/subnet_contract.rs
```diff
@@ -921,3 +921,98 @@ async fn test_subnet_send_token_processing(
     context.shutdown().await?;
     Ok(())
 }
+
+/// Test sync of blocks and generating certificates from genesis block
+#[rstest]
+#[test(tokio::test)]
+#[timeout(std::time::Duration::from_secs(600))]
+#[serial]
+async fn test_sync_from_genesis(
```
hadjiszs, line 930: Here we have
- should start creating from the first block if the source head fetched from the TCE is `0` or non-existant

Do we have also one test for the general case?
- should start creating from `N+1` if the source head fetched from the TCE is `N`

crates/topos-sequencer-subnet-runtime/tests/subnet_contract.rs
```diff
@@ -921,3 +921,98 @@ async fn test_subnet_send_token_processing(
...
+                    block_number, cert.id, cert.target_subnets
+                );
+                receieved_certificate_block_heights.push(block_number);
+
+                if receieved_certificate_block_heights
+                    .iter()
+                    .take(subnet_height as usize)
+                    .copied()
+                    .collect::<Vec<_>>()
+                    == expected_blocks
```
hadjiszs, line 994: In order to ensure that the two streams are without any shift: we may like checking one anchor as well, e.g., that the certificate 2 and block 2 are sharing the same state root hash or any other shared field

hadjiszs (dismissed):

crates/topos-sequencer-subnet-runtime/tests/subnet_contract.rs
atanmarko: It is. I will keep it for the second part of the test, sync from specific block (not genesis)

crates/topos-sequencer-subnet-runtime/tests/subnet_contract.rs
atanmarko: I'll expand the test to test for start from N+1

crates/topos-sequencer-subnet-runtime/src/proxy.rs
```diff
@@ -193,7 +194,7 @@ impl SubnetRuntimeProxy {
                             runtime_proxy.clone(),
                             &mut subnet_listener,
                             certification.clone(),
-                            next_block_number,
+                            next_block_number as u64,
```
gruberb, line 197: Why `u64` here and not `u128` like above (instead of signed)?

crates/topos-tce-proxy/src/worker.rs
```diff
@@ -69,7 +69,7 @@ impl TceProxyWorker {
 
         let source_last_certificate = if source_last_pending_certificate.is_none() {
             // There are no pending certificates on the TCE
-            // Block height to get next from subnet is position +1
+            // Block height to get next from subnet is position + 1
```
gruberb, line 72: Remove space again

gruberb (approved):

crates/topos-sequencer-subnet-runtime/src/proxy.rs
atanmarko: Above is i128, to indicate with -1 that we do not have any block so far. API for subnet client expects u64 in general.

crates/topos-sequencer-subnet-runtime/tests/subnet_contract.rs
hadjiszs: The test is now like >200 lines while it feels better suited to be separated in two tests instead no?

atanmarko:
@hadjiszs I'll merge this one, we need this genesis sync fix for the release. Any other change I'll implement in additional PR

PR #306, opened 2023-09-18
feat: cleaning up the synchronizer

Freyskeyd:
# Description

This PR aims to add and update passive checkpoint sync implementation and add tests on the storage part.

## PR Checklist:


crates/topos-tce-api/src/grpc/mod.rs
```diff
@@ -213,58 +215,38 @@ impl ApiService for TceGrpcService {
...
-                    "Can't get last pending certificates: {e}"
-                ))),
+        let last_pending_certificate = self
+            .store
+            .get_pending_certificates_for_subnets(&subnet_ids)
+            .map_err(|e| Status::internal(format!("Can't get last pending certificates: {e}")))?
+            .into_iter()
+            .map(|(subnet_id, (index, maybe_certificate))| {
+                (
+                    base64::engine::general_purpose::STANDARD.encode(subnet_id.as_array()),
```
dvdplm, line 231: So this line looked annoyingly long and I went to check the docs to see if there's not something less fugly, but no, this is how it's supposed to be used (!). I did some more digging and it seems like the `base64` maintainer is struggling to keep up maintenance. One notable feature that is missing is constant time encoding/decoding. 

What do you think about switching to the RustCrypto `base64ct` crate instead? I don't think being constant time matters in this particular code, but if we have a single crate in the code base that happens to be constant time then we save our future selves a papercut: no need to even think about "so do I need the constant time base64 crate here or is the other one ok?".

crates/topos-tce-api/src/runtime/builder.rs
```diff
@@ -90,6 +90,12 @@ impl RuntimeBuilder {
         let (api_event_sender, api_event_receiver) = mpsc::channel(2048);
 
         let (health_reporter, tce_status, grpc) = ServerBuilder::default()
+            .with_store(
+                self.store
+                    .clone()
+                    .take()
+                    .expect("Unable to build gRPC Server, Store is missing"),
```
dvdplm, line 97: Same comment as above.

dvdplm (approved):

crates/topos-tce-api/src/grpc/builder.rs
```diff
@@ -59,7 +67,15 @@ impl ServerBuilder {
             status: status.clone(),
         });
 
-        let service = ApiServiceServer::new(TceGrpcService { command_sender });
+        let store = self
+            .store
+            .take()
+            .expect("Cannot build GraphQL server without a FullNode store");
```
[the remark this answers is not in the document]
Freyskeyd: That's part of the refactor effort that I initiate, but it's not the scope of the PR to remove builders.

crates/topos-tce-api/src/grpc/mod.rs
Freyskeyd: Handled by another PR

PR #310, opened 2023-09-22
Documenting tce-proxy

dvdplm:
Added some docs while reading the tce-proxy code


crates/topos-sequencer/src/lib.rs
```diff
@@ -111,7 +111,7 @@ pub async fn launch(
     // TODO: Revise this approach?
     let (tce_proxy_worker, source_head_certificate_id) = match TceProxyWorker::new(TceProxyConfig {
         subnet_id,
-        base_tce_api_url: config.tce_grpc_endpoint.clone(),
+        tce_endpoint: config.tce_grpc_endpoint.clone(),
```
dvdplm, line 114: I can back out this change if it's annoying to mix doc changes with style.

Freyskeyd: I think at some point we'll remove the `tce` context as it gets less and less important

atanmarko (approved):

PR #311, opened 2023-09-26
feat: use constant-time Base64 encoding

dvdplm:
# Description

We already have the `base64ct` crate in the repo (dep of ethers I believe) and I think one is enough! It is also constant time, which is strictly better than the current `base64` and seem to have a better API.


crates/topos-api/tests/tce_layer.rs
```diff
@@ -76,7 +76,7 @@ async fn create_tce_layer() {
             let mut map = HashMap::new();
             for subnet_id in subnet_ids {
                 map.insert(
-                    base64::engine::general_purpose::STANDARD.encode(&subnet_id.value),
+                    Base64::encode_string(&subnet_id.value),
```
Freyskeyd, line 79:
```suggestion
                    subnet_id.to_string()
```

I think we could just use `to_string` here

Freyskeyd (approved):

crates/topos-api/tests/tce_layer.rs
dvdplm: Hmm, doesn't seem like that is possible. I just noticed that the `SubnetId` is not [this type](https://github.com/topos-protocol/topos/blob/d7122db7fa3ba6408c0951ec67d3a5333c33433b/crates/topos-uci/src/subnet_id.rs#L9) but rather the `prost` version ([this](https://github.com/topos-protocol/topos/blob/d7122db7fa3ba6408c0951ec67d3a5333c33433b/crates/topos-api/src/grpc/generated/topos.shared.v1.rs#L13)).

PR #313, opened 2023-09-26
fix: submit certificate to tce using backoff algorithm in case of error

atanmarko:
# Description

In case that `topos-tce-proxy` client fails to submit certificate to the TCE, it will retry using backoff algorith.

Fixes TP-719

## PR Checklist:


dvdplm:
@atanmarko Did you see [Tp-720](https://linear.app/toposware/issue/TP-720/as-sequencer-i-want-to-use-the-ethers-retrypolicy-instead-of-the)? Is this an intermediate step towards that?

Freyskeyd (approved):

PR #318, opened 2023-09-29
feat: add start block

atanmarko:
# Description

Add sequencer parameter `--start-block` to start syncing from particular subnet block.

Fixes TP-742

## PR Checklist:

### DIFF
--- a/crates/topos-sequencer-subnet-runtime/src/proxy.rs
+++ b/crates/topos-sequencer-subnet-runtime/src/proxy.rs
@@ -87,6 +87,7 @@ impl SubnetRuntimeProxy {
             None,
             config.verifier,
             signing_key.clone(),
+            config.start_block,
         )?;
 
         let runtime_proxy = Arc::new(Mutex::from(Self {
@@ -104,9 +105,14 @@ impl SubnetRuntimeProxy {
             let runtime_proxy = runtime_proxy.clone();
             let subnet_contract_address = subnet_contract_address.clone();
             tokio::spawn(async move {
-                let default_block_sync: i128 =
-                    config.start_block.map(|i| i as i128).unwrap_or(-1);
-                let mut latest_acquired_subnet_block_number: i128 = default_block_sync;
+                // if `start_block` parameter is provided, first block retrieved (genesis certificate)
+                // will be start block. `default_block_sync_start` is hence `start_block`-1
+                // as first retrieved block is latest_acquired_subnet_block_number + 1
+                let default_block_sync_start: i128 = config
+                    .start_block
+                    .map(|block_number| (block_number - 1) as i128)
+                    .unwrap_or(-1);
+                let mut latest_acquired_subnet_block_number: i128 = default_block_sync_start;
 
                 {
                     // To start producing certificates, we need to know latest delivered or pending certificate id from TCE
@@ -132,7 +138,7 @@ impl SubnetRuntimeProxy {
                                 let cert_id = certificate_and_position.map(|(id, _position)| id);
                                 let position: i128 = certificate_and_position
                                     .map(|(_id, position)| position as i128)
-                                    .unwrap_or(default_block_sync);
+                                    .unwrap_or(default_block_sync_start);
                                 // Certificate generation is now ready to run
                                 certification.last_certificate_id = cert_id;
                                 latest_acquired_subnet_block_number = position;
--- a/crates/topos-sequencer-subnet-runtime/src/proxy.rs
+++ b/crates/topos-sequencer-subnet-runtime/src/proxy.rs
@@ -105,9 +105,9 @@ impl SubnetRuntimeProxy {
             let runtime_proxy = runtime_proxy.clone();
             let subnet_contract_address = subnet_contract_address.clone();
             tokio::spawn(async move {
-                // if `start_block` parameter is provided, first block retrieved (genesis certificate)
-                // will be start block. `default_block_sync_start` is hence `start_block`-1
-                // as first retrieved block is latest_acquired_subnet_block_number + 1
+                // If the `start_block` sequencer parameter is provided, first block retrieved from blockchain (for genesis certificate)
+                // will be `start_block`. `default_block_sync_start` is hence `start_block`-1
+                // as first block retrieved from subnet node is `latest_acquired_subnet_block_number` + 1
                 let default_block_sync_start: i128 = config
                     .start_block
                     .map(|block_number| (block_number - 1) as i128)
--- a/crates/topos-sequencer-subnet-runtime/src/proxy.rs
+++ b/crates/topos-sequencer-subnet-runtime/src/proxy.rs
@@ -122,10 +122,10 @@ impl SubnetRuntimeProxy {
                                     "Source head certificate id received {:?}",
                                     certificate_and_position
                                 );
-                                // If tce source head position is provided, continue syncing from it
-                                // If `start_block` sequencer parameter is provided and tce source head is missing,
-                                // we should start syncing from that block instead of genesis
-                                // If both tce source head position and start_block parameter are not provided,
+                                // If tce source head position is provided, continue synchronizing from it
+                                // If the `start_block` sequencer parameter is provided and tce source head is missing,
+                                // we should start synchronizing from that block instead of genesis
+                                // If neither tce source head position nor start_block parameters are provided,
                                 // sync should start form -1, so that first fetched is subnet genesis block
                                 let default_block_sync: i128 =
                                     config.start_block.map(|i| i as i128).unwrap_or(-1);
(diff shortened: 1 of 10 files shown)


crates/topos-sequencer-subnet-runtime/src/proxy.rs
```diff
@@ -122,11 +122,17 @@ impl SubnetRuntimeProxy {
                                     "Source head certificate id received {:?}",
                                     certificate_and_position
                                 );
-                                // If the position is not provided, it should start form -1, so that first fetched is subnet genesis block
+                                // If tce source head position is provided, continue syncing from it
```
dvdplm, line 125:
```suggestion
                                // If tce source head position is provided, continue synching from it
```

crates/topos-sequencer-subnet-runtime/src/proxy.rs
```diff
@@ -122,11 +122,17 @@ impl SubnetRuntimeProxy {
...
+                                // If `start_block` sequencer parameter is provided and tce source head is missing,
```
dvdplm, line 126:
```suggestion
                                // If the `start-block` sequencer parameter is provided and tce source head is missing,
```

crates/topos-sequencer-subnet-runtime/src/proxy.rs
```diff
@@ -122,11 +122,17 @@ impl SubnetRuntimeProxy {
...
+                                // we should start syncing from that block instead of genesis
```
dvdplm, line 127:
```suggestion
                                // we should start synching from that block instead of genesis
```

crates/topos-sequencer-subnet-runtime/src/proxy.rs
```diff
@@ -122,11 +122,17 @@ impl SubnetRuntimeProxy {
...
+                                // If both tce source head position and start_block parameter are not provided,
```
dvdplm, line 128:
```suggestion
                                // If neither tce source head position nor `start-block` parameters are provided,
```

crates/topos-sequencer-subnet-runtime/src/proxy.rs
```diff
@@ -122,11 +122,17 @@ impl SubnetRuntimeProxy {
...
+                                // sync should start form -1, so that first fetched is subnet genesis block
+                                let default_block_sync: i128 =
```
dvdplm, line 130: `i128`? Why?

crates/topos/src/components/sequencer/commands/run.rs
```diff
@@ -48,6 +48,11 @@ pub struct Run {
     /// If not provided open telemetry will not be used
     #[arg(long, env = "TOPOS_OTLP_SERVICE_NAME")]
     pub otlp_service_name: Option<String>,
+
+    /// Start syncing from particular block number
+    /// Default is to sync from genesis block 0
```
dvdplm, line 53:
```suggestion
    /// Start synchronizing from particular block number.
    /// Default is to sync from genesis block (0).
```

crates/topos/src/config/sequencer.rs
```diff
@@ -32,6 +32,10 @@ pub struct SequencerConfig {
 
     /// OTLP service name, not used if not provided
     pub otlp_service_name: Option<String>,
+
+    /// Start syncing from particular block number
+    /// Default is to sync from genesis block 0
```
dvdplm, line 37:
```suggestion
    /// Start synchronizing from particular block number.
    /// Default is to sync from genesis block (0).
```

crates/topos/tests/snapshots/sequencer__sequencer_help_display.snap
```diff
@@ -27,6 +27,8 @@ Options:
           Socket of the opentelemetry agent endpoint If not provided open telemetry will not be used [env: TOPOS_OTLP_AGENT=]
       --otlp-service-name <OTLP_SERVICE_NAME>
           Otlp service name If not provided open telemetry will not be used [env: TOPOS_OTLP_SERVICE_NAME=]
+      --start-block <START_BLOCK>
+          Start syncing from particular block number Default is to sync from genesis block 0 [env: TOPOS_START_BLOCK=]
```
dvdplm, line 31:
```suggestion
          Start synching from particular block number Default is to sync from genesis block 0 [env: TOPOS_START_BLOCK=]
```

dvdplm (approved):
lgtm modulo spelling

crates/topos-sequencer-subnet-runtime/src/proxy.rs
Freyskeyd: I feel that `synching` isn't really standard in computer industry. I would prefer sticking to `syncing` or even `synchronizing`

dvdplm: Both spellings are correct-ish. Let's stick to "synchronizing" then.

crates/topos-sequencer-subnet-runtime/src/proxy.rs
```diff
@@ -122,11 +122,17 @@ impl SubnetRuntimeProxy {
...
+                                    config.start_block.map(|i| i as i128).unwrap_or(-1);
```
hadjiszs, line 131: Better to move it one scope above in order to set `latest_acquired_subnet_block_number` [here](https://github.com/topos-protocol/topos/pull/318/files#diff-7f258e647b37319822e9ed2541297b6800806a063fa7667e60da9806ef72e891R107) as well which is still set to `-1`

crates/topos-sequencer-subnet-runtime/src/proxy.rs
atanmarko: Because I can not indicate unknown (default starting) block with `u64`

crates/topos-sequencer-subnet-runtime/src/proxy.rs
atanmarko: Currently it does not change flow (it would panic if source head could not be retrieved), but I agree, it should be moved up

crates/topos-sequencer-subnet-runtime/src/proxy.rs
```diff
@@ -104,7 +105,14 @@ impl SubnetRuntimeProxy {
             let runtime_proxy = runtime_proxy.clone();
             let subnet_contract_address = subnet_contract_address.clone();
             tokio::spawn(async move {
-                let mut latest_acquired_subnet_block_number: i128 = -1;
+                // if `start_block` parameter is provided, first block retrieved (genesis certificate)
+                // will be start block. `default_block_sync_start` is hence `start_block`-1
+                // as first retrieved block is latest_acquired_subnet_block_number + 1
```
dvdplm, line 110:
```suggestion
                // If the `start_block` parameter is provided, the first block retrieved (genesis certificate)
                // will be the start block. `default_block_sync_start` is hence `start_block`-1
                // as first retrieved block is latest_acquired_subnet_block_number + 1
```

crates/topos-sequencer-subnet-runtime/src/proxy.rs
atanmarko: @dvdplm Block number is always positive u64. Last block number that sequencer previously retrieved from subnet node could be 7894 or could be 0 (we have genesis block), and then next block that should be read from subnet node is `latest_acquired_subnet_block_number` + 1. If we have a new subnet blockchain and we have no information about it we should read first genesis block 0. To indicate that we say that `latest_acquired_subnet_block_number` is invalid (value -1) so we can read genesis block as `latest_acquired_subnet_block_number`+1

atanmarko:
@dvdplm 

```
Hmm, I do not follow. Can you explain, please?
```

Please check my comment above about `latest_acquired_subnet_block_number`. `Start_block` is command line parameter that says that genesis certificate is not block zero but block number given with command line parameter `start_block`.

Here is updated comment in code:
```
                // If the `start_block` sequencer parameter is provided, first block retrieved (genesis certificate)
                // will be `start_block`. `default_block_sync_start` is hence `start_block`-1
                // as first retrieved block is latest_acquired_subnet_block_number + 1
```

crates/topos-sequencer-subnet-runtime/src/proxy.rs
dvdplm: I see. Not super elegant. :/

atanmarko: What would you do as alternative? I don't like the idea of introducing another Option to mark  if we have initial info about genesis block and handling everywhere through both cases for block number

dvdplm: Good question. I'll have a think and see what I come up with.

PR #319, opened 2023-10-02
feat: replace libp2p peer ids with validator ids

gruberb:
# Description

Replacing `PeerId` from `libp2p` in the broadcast layer with our own `ValidatorId` (which is a Ethereum public address -`H160` as for now). 

Fixes TP-708

## Additions and Changes

* The broadcast layer stores the network peers now as `HashSet<ValidatorId>`, and supplies the list of validators through the `genesis.json` file during start.
* As soon as the node starts, `stable_sample` is set to `true` right away
* The `cert_delivery` test is adjusted so each node has `validators` and a `validator_id` to start with

# Breaking changes

IMPORTANT: This removes the `--push-peer-list` command AND api endpoint. 

## PR Checklist:


crates/topos-tce-api/Cargo.toml
```diff
@@ -12,6 +12,7 @@ topos-api = { path = "../topos-api", features = ["graphql"] }
 topos-core = { workspace = true, features = ["uci", "api"] }
 topos-metrics = { path = "../topos-metrics" }
 topos-tce-storage = { path = "../topos-tce-storage" }
+topos-tce-transport = { path = "../topos-tce-transport" }
```
Freyskeyd, line 15: Why is it needed?

crates/topos-tce-api/src/grpc/builder.rs
```diff
@@ -63,7 +63,6 @@ impl ServerBuilder {
         }));
 
         let console = ConsoleServiceServer::new(TceConsoleService {
-            command_sender: command_sender.clone(),
```
Freyskeyd, line 66: maybe we can keep this as we may add more command in the future to manage a node.

crates/topos-tce-gatekeeper/Cargo.toml
```diff
@@ -13,9 +13,10 @@ rand.workspace = true
 thiserror.workspace = true
 tracing.workspace = true
 tokio = { workspace = true, features = ["full"] }
-
+libp2p.workspace = true
```
Freyskeyd, line 16:
```suggestion
```

crates/topos-tce-gatekeeper/src/builder.rs
```diff
@@ -1,8 +1,8 @@
 use std::future::IntoFuture;
 
 use futures::{future::BoxFuture, FutureExt};
+use libp2p::PeerId;
```
Freyskeyd, line 4:
```suggestion
```

crates/topos-tce-gatekeeper/src/client.rs
```diff
@@ -1,10 +1,8 @@
-use crate::{
-    GatekeeperCommand, GatekeeperError, GetAllPeers, GetAllSubnets, GetRandomPeers, PushPeerList,
-};
+use crate::{GatekeeperCommand, GatekeeperError, GetAllPeers, GetAllSubnets, GetRandomPeers};
 use async_trait::async_trait;
+use libp2p::PeerId;
```
Freyskeyd, line 3:
```suggestion
```

crates/topos-tce-gatekeeper/src/lib.rs
```diff
@@ -17,9 +17,9 @@ mod tests;
 
 pub use client::Client;
 pub use client::GatekeeperClient;
+use libp2p::PeerId;
```
Freyskeyd, line 20:
```suggestion
```

crates/topos-tce-transport/src/lib.rs
```diff
@@ -5,8 +5,51 @@ use serde::{Deserialize, Serialize};
...
+    }
+}
+
+impl From<H160> for ValidatorId {
+    fn from(address: H160) -> Self {
+        ValidatorId(address)
+    }
+}
+
+impl TryFrom<&str> for ValidatorId {
```
Freyskeyd, line 38: Can be replace by [FromStr](https://doc.rust-lang.org/std/str/trait.FromStr.html) implementation

crates/topos-tce/src/app_context/api.rs
```diff
@@ -45,39 +42,6 @@ impl AppContext {
...
-                                "Gatekeeper has detected changes on the peer list, new sample in \
-                                 creation"
-                            );
-                            if sampler.peer_changed(peers).await.is_err() {
-                                _ = sender.send(Err(RuntimeError::UnableToPushPeerList));
-                            } else {
-                                api.set_active_sample(true).await;
-                                if events.send(Events::StableSample).await.is_err() {
-                                    error!("Unable to send StableSample event");
-                                }
```
Freyskeyd, line 67: You need to update the `active_sample` to make the service healthy and ready.
You can put it directly as true when the node starts

crates/topos-tce-transport/src/lib.rs
gruberb: The last PR explicitly wanted a `TryFrom` impl instead a `FromStr` :)

Freyskeyd: Mmh no it was switching from `From` to `TryFrom`.
`FromStr` is returning a `Result` to catch errors.

tools/node_config/subnet/topos/genesis.json
```diff
@@ -0,0 +1,51 @@
...
+        "epochSize": 30000,
+        "type": "PoA",
+        "validator_type": "bls"
+      }
+    },
+    "blockGasTarget": 0,
+    "burnContract": null,
+    "burnContractDestinationAddress": "0x0000000000000000000000000000000000000000"
+  },
+  "bootnodes": [
```
gruberb, line 45: Even though the tests run through, there is some tweaking to be done around the `bootnodes`. I assume we need to adjust this to just one `bootnode`, and maybe give it a fixed `PeerId` in the `init.sh`, and replace the list here. cc: @Freyskeyd

crates/topos-crypto/src/messages.rs
```diff
@@ -16,7 +16,7 @@ pub enum MessageSignerError {
 #[derive(Debug)]
 pub struct MessageSigner {
     pub public_address: Address,
-    wallet: LocalWallet,
+    pub wallet: LocalWallet,
```
dvdplm, line 19: Let's document all new/changed `pub` items. ;)

crates/topos-tce-api/src/grpc/builder.rs
```diff
@@ -59,7 +59,7 @@ impl ServerBuilder {
             .expect("Cannot build gRPC without an InternalRuntimeCommand sender");
 
         let status = Arc::new(RwLock::new(StatusResponse {
-            has_active_sample: false,
+            has_active_sample: true,
```
dvdplm, line 62: What does this do?

crates/topos-tce-api/src/grpc/console.rs
```diff
@@ -1,60 +1,21 @@
...
-    console_service_server::ConsoleService, PushPeerListRequest, PushPeerListResponse,
-    StatusRequest, StatusResponse,
+    console_service_server::ConsoleService, StatusRequest, StatusResponse,
 };
-use topos_p2p::PeerId;
-
-use crate::runtime::InternalRuntimeCommand;
 
 pub(crate) struct TceConsoleService {
+    #[allow(dead_code)]
```
dvdplm, line 12: Can we have a team rule that all uses `#[allow(dead_code)]` comes with a comment justifying it? Like `// Will be wired up as part of TP-xxx` or whatever happens to be the case.

crates/topos-tce-broadcast/src/double_echo/mod.rs
```diff
@@ -128,27 +129,16 @@ impl DoubleEcho {
...
-            forwarding_subscriptions_receiver,
-            task_manager_message_receiver,
-        );
+        let mut task_completion = self.spawn_task_manager(task_manager_message_receiver);
 
         info!("DoubleEcho started");
 
         let shutdowned: Option<oneshot::Sender<()>> = loop {
             tokio::select! {
                 biased;
```
dvdplm, line 140: What does this do?

crates/topos-test-sdk/src/tce/gatekeeper.rs
```diff
@@ -9,9 +9,11 @@ use topos_tce_gatekeeper::{Client, GatekeeperError};
 
 pub async fn create_gatekeeper<P: Into<PeerId>>(
     peer_id: P,
+    peer_list: Vec<PeerId>,
```
dvdplm, line 12: Shouldn't this be generic over `P` too? `peerlist: Vec<P>` (or remove the generic altogether, which would be even better if possible)

crates/topos-tce-transport/src/lib.rs
```diff
@@ -5,8 +5,51 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Debug, Error)]
+pub enum ValidatorIdConversionError {
+    #[error("Failed to parse address string as H160")]
+    ParseError,
+    #[error("Failed to convert byte array into H160")]
+    InvalidByteLength,
+}
+
+#[derive(Clone, Copy, Default, Debug, Serialize, Deserialize, Eq, PartialEq, Hash)]
+pub struct ValidatorId(H160);
```
Freyskeyd, line 20: Maybe the `ValidatorId` should be part of the `topos-core` as it'll be used by other component not just the `broadcast`

crates/topos-tce-gatekeeper/Cargo.toml
```diff
@@ -13,9 +13,9 @@ rand.workspace = true
 thiserror.workspace = true
 tracing.workspace = true
 tokio = { workspace = true, features = ["full"] }
-
 topos-core = { workspace = true, features = ["uci"] }
 topos-p2p = { path = "../topos-p2p" }
+topos-tce-transport = { path = "../topos-tce-transport" }
```
Freyskeyd, line 18: Do we really need this dep?

tools/init.sh
```diff
@@ -108,6 +114,23 @@ case "$1" in
            echo "PEER_ID: $PEER"
            echo "TCE_LOCAL_KS: $HOSTNAME"
 
+           INDEX=0
+
+           # For libp2p_keys.json
+           LIBP2P_KEY=$(jq -r ".keys[$INDEX]" /tmp/shared/libp2p_keys.json)
+           echo -n "$LIBP2P_KEY" > $TOPOS_HOME/node/test/libp2p/libp2p.key
+           jq "del(.keys[$INDEX])" /tmp/shared/libp2p_keys.json > /tmp/shared/libp2p_keys_temp.json && mv /tmp/shared/libp2p_keys_temp.json /tmp/shared/libp2p_keys.json
```
Freyskeyd, line 122: If the nodes are spawn at the same time, they'll all have the `INDEX=0` key right?

maybe we can use a random index like:

```
INDEX=$(jq --arg i $(($RANDOM % $(jq length  /tmp/shared/libp2p_keys.json))) '$i|tonumber'  /tmp/shared/libp2p_keys.json)
```

tools/node_config/node/test/config.toml
```diff
@@ -0,0 +1,13 @@
...
+subnet-id = "topos"
+
+[edge]
+
+[tce]
+db-path = "node_config/node/test/./tce_rocksdb"
+graphql-api-addr = "0.0.0.1:4030"
+grpc-api-addr = "0.0.0.1:1340"
+libp2p-api-addr = "0.0.0.1:9090"
+metrics-api-addr = "0.0.0.1:3000"
```
Freyskeyd, line 13: Break line at the end

tools/docker-compose.yml
```diff
@@ -3,14 +3,14 @@ services:
...
     image: ghcr.io/topos-protocol/topos:main
     init: true
     labels:
       "autoheal": "true"
       "prometheus-job": "boot"
     healthcheck:
-      test: ./topos tce push-peer-list --node http://localhost:1340 --format json /tmp/shared/peer_ids.json && ./topos tce status --node http://localhost:1340
+      test: ./topos tce status --node http://localhost:1340
       interval: 15s
     volumes:
```
Freyskeyd, line 15: You should add the key files here directly

Freyskeyd (changes requested):

tools/init.sh
gruberb: Since this is blocking, we can always have the same `index`. We lock the file and access it, write it back, release the lock. So every peer just always takes the first from the array without a problem.

tools/docker-compose.yml
gruberb: What exactly do you mean with key files?

crates/topos-tce-broadcast/src/double_echo/mod.rs
gruberb: Every receiving message to a node is getting put on to a `TaskManager`, and is being worked on in the background. Here we start the `TaskManager`, and give the receiver end of a channel. The `double_echo` is receiving messages, and the `TaskManager` is receiving them, spawning a new task, or feeding messages to an existing task.

One certificate equals a task.

dvdplm: No I meant specifically this line, `biased;` – I'm not very good with async code in general and the `select!` macro is tricky for me sometimes.

(Good explanation though!)

tools/docker-compose.yml
Freyskeyd: I mean all the files that you added in the `Dockerfile` should be part of the `docker-compose`

Dockerfile
```diff
@@ -49,5 +49,14 @@ RUN apt-get update && apt-get install -y \
...
 RUN mkdir /tmp/shared
+RUN mkdir -p /tmp/node_config/subnet/topos
+RUN mkdir -p /tmp/node_config/node/test/consensus
+RUN mkdir -p /tmp/node_config/node/test/libp2p
+
+COPY tools/node_config/node/test/config.toml /tmp/node_config/node/test/config.toml
+COPY tools/node_config/subnet/topos/genesis.json /tmp/node_config/subnet/topos/genesis.json
+COPY tools/libp2p_keys.json /tmp/shared/libp2p_keys.json
+COPY tools/validator_bls_keys.json /tmp/shared/validator_bls_keys.json
+COPY tools/validator_keys.json /tmp/shared/validator_keys.json
```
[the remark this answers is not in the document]
Freyskeyd: Indeed, do we want to have tests files in the `Dockerfile` that can end up in production?

The `Dockerfile` should only contain file that are needed for the creation of the binary, all files and configuration used for runtime should be added via volume when running the container.

tools/init.sh
Freyskeyd: Which part are you defining as blocking?
Reading a file doesn't create a lock and writing to it either because it's a shared file, that's why we end up having lock mechanisms for the creation of the peers.

tools/docker-compose.yml
```diff
@@ -62,17 +62,17 @@ services:
...
       - env/node.env
       - env/telemetry.env
     environment:
-      - RUST_LOG=topos_tce_storage=info,topos_tce_synchronizer=debug
+      - RUST_LOG=topos=debug,topos_tce_storage=info,topos_tce_synchronizer=debug
 
   sync:
-    image: ghcr.io/topos-protocol/topos:main
-    command: sync tce run
+    image: 4c7598914c59
```
hadjiszs, line 74: relicate?

hadjiszs (changes requested):

crates/topos-tce-broadcast/src/double_echo/mod.rs
gruberb: This will cause the `Futures` to be selected in order:
https://docs.rs/tokio/latest/tokio/macro.select.html#fairness

Dockerfile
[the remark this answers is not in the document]
gruberb: Fair enough. My thinking was that we just use Docker for the local test setup, right? I will change it though.

tools/init.sh
gruberb: Ups, exactly. I had a `lock`  version locally as an experiment and somehow a different version made it upstream. My bad. Yeah, either using a lock like with the `peer_list` or using a random index 👍

crates/topos-tce/src/lib.rs
```diff
@@ -70,6 +70,8 @@ pub async fn run(
     let mut boot_peers = config.boot_peers.clone();
     boot_peers.retain(|(p, _)| *p != peer_id);
 
+    let peer_list = boot_peers.iter().map(|(p, _)| *p).collect::<Vec<_>>();
```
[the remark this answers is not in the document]
gruberb: To be honest, I didn't come up with a better solution 🤷

crates/topos-tce-broadcast/src/tests/mod.rs
```diff
@@ -50,26 +51,34 @@ struct Context {
...
+    let mut rng = rand::thread_rng();
+
     let message_signer = Arc::new(MessageSigner::from_str(PRIVATE_KEY).unwrap());
 
     let mut validators = HashSet::new();
     let validator_id = ValidatorId::from(message_signer.public_address);
     validators.insert(validator_id);
 
+    for _ in 1..params.nb_peers {
+        let private_key: [u8; 32] = rng.gen();
```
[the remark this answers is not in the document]
gruberb: The one in the `topos-test-sdk` is also creating a `HashSet` of `message_signers`. So creating a `util` function is different enough that for `tests`, I rather have code at hand I can read through and understand immedietaly, without having to understand function parameters (which we would need to say "Don't create  `HashSet` of `message_signers` in this case. 

So yeah, in non-test code, we would find a better way most likely, but for tests? I don't see the benefit for overcomplicating it.

crates/topos-tce/src/lib.rs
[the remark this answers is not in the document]
dvdplm: Ah, you need both `boot_peers` and `peer_list`. Hmm, ok, well, then I don't think there's much to be done.

dvdplm (approved):
lgtm

crates/topos-tce-proxy/tests/tce_tests.rs
```diff
@@ -349,6 +349,7 @@ async fn test_tce_open_stream_with_checkpoint(
     #[future]
     start_node: TceContext,
 ) -> Result<(), Box<dyn std::error::Error>> {
+    println!("HEERE");
```
hadjiszs, line 352:
```suggestion
```

hadjiszs (approved):

Freyskeyd (changes requested):

Dockerfile
```diff
@@ -49,5 +49,8 @@ RUN apt-get update && apt-get install -y \
 USER topos:topos
 
 RUN mkdir /tmp/shared
+RUN mkdir -p /tmp/node_config/subnet/topos
+RUN mkdir -p /tmp/node_config/node/test/consensus
+RUN mkdir -p /tmp/node_config/node/test/libp2p
```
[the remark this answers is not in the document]
gruberb: I tried, and no, I need to create the folders or it won't have it during the `init.sh` script and it fails.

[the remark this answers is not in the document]
Freyskeyd: Then just create the topos home in the home folder and don't recreate the whole directory structure. I would prefer not to have complete tree in docker as it can change.

PR #322, opened 2023-10-03
chore: rename `genesis.json` used in tests to `genesis-example.json`

hadjiszs:
# Description

Simply renaming the `genesis.json` file used in the tests to `genesis-example.json` to avoid confusions


atanmarko (approved):

dvdplm (approved):
Can we also clear out the file of things like e.g. the bootnodes that can't be used anyway?

hadjiszs:
> Can we also clear out the file of things like **e.g. the bootnodes that can't be used anyway**?

ah they do actually, the tests are also covering the parsing of this section

dvdplm:
> ah they do actually, the tests are also covering the parsing of this sect

Can we have a copy just for the tests then? `genesis-tests.json`?

PR #323, opened 2023-10-04
fix: tce private key encoding

atanmarko:
# Description

Fix encoding of the TCE private key on startup.

Raw secret keys were converted to string using utf8 convert function which failed occasionally because keys are not valid utf8. Also, wallets used expect hex encoded raw key bytes, not utf8 character values of secret key bytes.

Fixes TP-746

## PR Checklist:

### DIFF
--- a/crates/topos-tce/src/lib.rs
+++ b/crates/topos-tce/src/lib.rs
@@ -48,9 +48,7 @@ pub async fn run(
 
     let message_signer = match &config.signing_key {
         Some(AuthKey::PrivateKey(pk)) => {
-            let bytes = pk.to_vec();
-            let bytes_str = std::str::from_utf8(&bytes)?;
-            MessageSigner::new(bytes_str)?
+            MessageSigner::new(&hex::encode(pk))?
         }
         _ => return Err(Box::try_from("Error, no singing key".to_string()).unwrap()),
     };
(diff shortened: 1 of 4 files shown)


crates/topos-tce/src/lib.rs
```diff
@@ -47,11 +47,7 @@ pub async fn run(
     };
 
     let message_signer = match &config.signing_key {
-        Some(AuthKey::PrivateKey(pk)) => {
-            let bytes = pk.to_vec();
-            let bytes_str = std::str::from_utf8(&bytes)?;
-            MessageSigner::new(bytes_str)?
-        }
+        Some(AuthKey::PrivateKey(pk)) => MessageSigner::new(&hex::encode(pk))?,
```
dvdplm, line 50: Was this the UTF8 related error we saw at start up?

dvdplm (approved):
lgtm

(but the description and associated linear ticket is not enough to understand *why* these code changes are here)

crates/topos-tce/src/lib.rs
atanmarko: Yes seems so

atanmarko:
> (but the description and associated linear ticket is not enough to understand _why_ these code changes are here)

Updated descrption.

gruberb (approved):
👍

atanmarko:
Closing as duplicate of https://github.com/topos-protocol/topos/pull/324

PR #324, opened 2023-10-04
fix: private key parsing

Freyskeyd:
# Description

This PR is fixing an issue while trying to parse the private key into a `LocalWallet`, instead of trying to parse `bytes` into `ut8 str` I directly use the `bytes` to generate the `wallet`.

## PR Checklist:

### DIFF
--- a/crates/topos-crypto/tests/messages.rs
+++ b/crates/topos-crypto/tests/messages.rs
@@ -1,3 +1,5 @@
+use std::str::FromStr;
+
 use rstest::*;
 use topos_core::uci::CertificateId;
 use topos_crypto::messages::MessageSigner;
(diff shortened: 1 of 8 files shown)


dvdplm (commented):
This seems like a different take on #323.

I think this one is better.

atanmarko:
Duplicate of https://github.com/topos-protocol/topos/pull/323

atanmarko:
> This seems like a different take on #323.
> 
> I think this one is better.

Ok lets then take this one

crates/topos-crypto/tests/messages.rs
```diff
@@ -20,7 +20,7 @@ pub fn test_signing_messages() {
         .expect("Cannot create Signature");
 
     let message_signer_receiver =
-        MessageSigner::new("a2e33a9bad88f7b7568228f51d5274c471a9217162d46f1533b6a290f0be1baf")
+        MessageSigner::new(b"a2e33a9bad88f7b7568228f51d5274c471a9217162d46f1533b6a290f0be1baf")
```
atanmarko, line 23: This is not valid. `a2` should be byte 162, here Rust parses it as two bytes 97, 50
Is this really what you wanted https://doc.rust-lang.org/reference/tokens.html#byte-string-literals

atanmarko: You can use instead:
```
hex::decode(""a2e33a9bad88f7b7568228f51d5274c471a9217162d46f1533b6a290f0be1baf").unwrap()
```

atanmarko:
> This seems like a different take on #323.
> 
> I think this one is better.

I think it is cleaner for testing to stay with string representation of keys

atanmarko:
@Freyskeyd @dvdplm This PR grows unnecessary complex. Could we switch back to much simpler variant #323?

atanmarko (approved):

PR #325, opened 2023-10-05
fix: extract validator pubkeys from genesis

dvdplm:
Refactor extraction of validators from the genesis file.


gruberb:
Is there a reasoning behind this refactor?

gruberb (approved):
LGTM, although personally I like the previous version, it's slightly easier to follow and understand.

JDawg287 (approved):

dvdplm:
> Is there a reasoning behind this refactor?

Not really, nothing beyond me reading code trying to fix a different problem and fixing a few warts and unneeded allocations.

PR #327, opened 2023-10-05
feat: adding GraphQL Subscription

Freyskeyd:
# Description

This PR is adding `GraphQL` subscriptions to the current API implementation. It introduces the basic functionality and creates the new `TransientStream` struct to manage those new streams.

## PR Checklist:

### DIFF
--- a/crates/topos-tce-api/src/graphql/query.rs
+++ b/crates/topos-tce-api/src/graphql/query.rs
@@ -113,6 +113,30 @@ impl QueryRoot {
 
 pub struct SubscriptionRoot;
 
+impl SubscriptionRoot {
+    /// Try to create a new [`Stream`] of delivered [`Certificate`]s to be used in a GraphQL subscription.
+    pub(crate) async fn new_transient_stream(
+        &self,
+        register: &mpsc::Sender<InternalRuntimeCommand>,
+    ) -> Result<impl Stream<Item = Certificate>, GraphQLServerError> {
+        let (sender, receiver) = oneshot::channel();
+        _ = register
+            .send(InternalRuntimeCommand::NewTransientStream { sender })
+            .await;
+
+        let stream: TransientStream = receiver
+            .await
+            .map_err(|_| {
+                GraphQLServerError::InternalError(
+                    "Communication error trying to create a new transient stream",
+                )
+            })?
+            .map_err(|e| GraphQLServerError::TransientStream(e.to_string()))?;
+
+        Ok(stream.map(|c| c.into()))
+    }
+}
+
 #[Subscription]
 impl SubscriptionRoot {
     /// This endpoint is used to received delivered certificates.
@@ -133,13 +157,6 @@ impl SubscriptionRoot {
                 GraphQLServerError::ParseDataConnector
             })?;
 
-        let (sender, receiver) = oneshot::channel();
-        _ = register
-            .send(InternalRuntimeCommand::NewTransientStream { sender })
-            .await;
-
-        let stream: TransientStream = receiver.await.unwrap().unwrap();
-
-        Ok(stream.map(|c| c.into()))
+        self.new_transient_stream(register).await
     }
 }
--- a/crates/topos-tce-api/src/graphql/query.rs
+++ b/crates/topos-tce-api/src/graphql/query.rs
@@ -1,7 +1,9 @@
 use std::sync::Arc;
 
-use async_graphql::{Context, EmptyMutation, EmptySubscription, Object, Schema};
+use async_graphql::{Context, EmptyMutation, EmptySubscription, Object, Schema, Subscription};
 use async_trait::async_trait;
+use futures::{Stream, StreamExt};
+use tokio::sync::{mpsc, oneshot};
 use topos_api::graphql::errors::GraphQLServerError;
 use topos_api::graphql::{
     certificate::{Certificate, CertificateId},
@@ -13,6 +15,9 @@ use topos_tce_storage::fullnode::FullNodeStore;
 use topos_tce_storage::store::ReadStore;
 
 use tracing::debug;
+
+use crate::runtime::InternalRuntimeCommand;
+use crate::stream::TransientStream;
 
 pub struct QueryRoot;
 pub(crate) type ServiceSchema = Schema<QueryRoot, EmptyMutation, EmptySubscription>;
@@ -103,5 +108,38 @@ impl QueryRoot {
         certificate_id: CertificateId,
     ) -> Result<Certificate, GraphQLServerError> {
         Self::certificate_by_id(ctx, certificate_id).await
+    }
+}
+
+pub struct SubscriptionRoot;
+
+#[Subscription]
+impl SubscriptionRoot {
+    /// This endpoint is used to received delivered certificates.
+    /// It uses a transient stream, which is a stream that is only valid for the current connection.
+    ///
+    /// Closing the connection will close the stream.
+    /// Starting a new connection will start a new stream and the client will not receive
+    /// any certificates that were delivered before the connection was started.
+    async fn watch_delivered_certificates(
+        &self,
+        ctx: &Context<'_>,
+    ) -> Result<impl Stream<Item = Certificate>, GraphQLServerError> {
+        let register = ctx
+            .data::<mpsc::Sender<InternalRuntimeCommand>>()
+            .map_err(|_| {
+                tracing::error!("Failed to get the transient register client from context");
+
+                GraphQLServerError::ParseDataConnector
+            })?;
+
+        let (sender, receiver) = oneshot::channel();
+        _ = register
+            .send(InternalRuntimeCommand::NewTransientStream { sender })
+            .await;
+
+        let stream: TransientStream = receiver.await.unwrap().unwrap();
+
+        Ok(stream.map(|c| c.into()))
     }
 }
--- a/crates/topos-tce-api/src/graphql/tests.rs
+++ b/crates/topos-tce-api/src/graphql/tests.rs
@@ -1,0 +1,122 @@
+use std::time::Duration;
+
+use crate::{
+    graphql::query::{QueryRoot, SubscriptionRoot},
+    runtime::InternalRuntimeCommand,
+    stream::TransientStream,
+};
+use async_graphql::{http, value, EmptyMutation, Schema};
+use futures::{SinkExt, StreamExt};
+use rstest::rstest;
+use test_log::test;
+use tokio::sync::{mpsc, oneshot};
+use uuid::Uuid;
+
+#[rstest]
+#[test(tokio::test)]
+#[timeout(Duration::from_secs(2))]
+async fn requesting_transient_stream_from_graphql() {
+    let (sender, mut receiver) = mpsc::channel(1);
+
+    tokio::spawn(async move {
+        let mut v = Vec::new();
+        while let Some(query) = receiver.recv().await {
+            if let InternalRuntimeCommand::NewTransientStream { sender } = query {
+                let (notifier, notifier_receiver) = oneshot::channel();
+                v.push(notifier_receiver);
+
+                let (_s, inner) = mpsc::channel(10);
+                _ = sender.send(Ok(TransientStream {
+                    stream_id: Uuid::new_v4(),
+                    notifier: Some(notifier),
+                    inner,
+                }));
+            }
+        }
+    });
+
+    let root = SubscriptionRoot {};
+
+    let result = root.new_transient_stream(&sender).await;
+
+    assert!(result.is_ok());
+}
+
+#[rstest]
+#[timeout(Duration::from_secs(4))]
+#[test(tokio::test)]
+async fn open_watch_certificate_delivered() {
+    let (mut tx, rx) = futures::channel::mpsc::unbounded();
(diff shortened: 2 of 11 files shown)


crates/topos-tce-api/src/graphql/query.rs
```diff
@@ -94,6 +99,7 @@ impl QueryRoot {
         from_source_checkpoint: SourceCheckpoint,
         first: usize,
     ) -> Result<Vec<Certificate>, GraphQLServerError> {
+        info!("In certificates");
```
dvdplm, line 102: left-over?

crates/topos-tce-api/src/graphql/query.rs
```diff
@@ -102,6 +108,49 @@ impl QueryRoot {
         ctx: &Context<'_>,
         certificate_id: CertificateId,
     ) -> Result<Certificate, GraphQLServerError> {
+        info!("In certificate");
```
dvdplm, line 111: left-over?

crates/topos-tce-api/src/graphql/query.rs
```diff
@@ -102,6 +108,49 @@ impl QueryRoot {
...
         Self::certificate_by_id(ctx, certificate_id).await
     }
 }
+
+pub struct SubscriptionRoot;
+
+impl SubscriptionRoot {
+    pub(crate) async fn new_transient_stream(
```
dvdplm, line 119: Docs?

I haven't read the full patch, so perhaps it'll be answered below, but what is a "transient stream" in this concept? What is the role of a `SubscriptionRoot`?

Q: this call looks like it doesn't really need to be async. I guess it's a general async rust question: should we strive to make as much code as possible follow the oneshot channel pattern seen here, or only where it's strictly needed/useful? Like, what is the point of sending the `InternalRuntimeCommand::NewTransientStream` and then wait for it, rather than do the work synchronously? (Honest question, not intended as a critique)

crates/topos-tce-api/src/graphql/query.rs
```diff
@@ -102,6 +108,49 @@ impl QueryRoot {
...
+        &self,
+        register: &mpsc::Sender<InternalRuntimeCommand>,
+    ) -> Result<impl Stream<Item = Certificate>, GraphQLServerError> {
+        let (sender, receiver) = oneshot::channel();
+        _ = register
+            .send(InternalRuntimeCommand::NewTransientStream { sender })
+            .await;
+
+        let stream: TransientStream = receiver.await.unwrap().unwrap();
```
dvdplm, line 128: unwrap bad.

crates/topos-tce-api/src/runtime/mod.rs
```diff
@@ -58,6 +59,7 @@ pub struct Runtime {
     pub(crate) broadcast_stream: broadcast::Receiver<CertificateDeliveredWithPositions>,
 
     pub(crate) storage: StorageClient,
+    pub(crate) transient_stream: HashMap<Uuid, Sender<Certificate>>,
```
dvdplm, line 62: Could this be a tuple perhaps?

crates/topos-tce-api/src/runtime/mod.rs
```diff
@@ -222,6 +225,36 @@ impl Runtime {
...
+
+                if sender
+                    .send(Ok(TransientStream {
+                        stream_id,
+                        inner: receiver,
+                        notifier: Some(shutdown),
+                    }))
+                    .is_err()
+                {
+                    error!("Unable to send new TransientStream");
```
dvdplm, line 253: Print the error too perhaps?

crates/topos-tce-api/src/stream/mod.rs
```diff
@@ -31,6 +31,38 @@ pub use self::commands::StreamCommand;
 pub use self::errors::StreamError;
 pub(crate) use self::errors::{HandshakeError, StreamErrorKind};
 
+/// [`TransientStream`] is a stream that live as long as the connection is open.
+/// A [`TransientStream`] will not receive any certificates that were delivered
+/// before the stream was ready to listen.
+///
+/// [`TransientStream`] is implementing [`futures::Stream`] and use a custom [`Drop`]
```
dvdplm, line 38:
```suggestion
/// [`TransientStream`] implements [`futures::Stream`] and use a custom [`Drop`]
```

crates/topos-tce-api/src/stream/mod.rs
```diff
@@ -31,6 +31,38 @@ pub use self::commands::StreamCommand;
...
+/// implementation to notify the `runtime` when ended.
+#[derive(Debug)]
+pub struct TransientStream {
+    pub(crate) inner: mpsc::Receiver<Certificate>,
+    pub(crate) stream_id: Uuid,
+    pub(crate) notifier: Option<oneshot::Sender<Uuid>>,
```
dvdplm, line 44: What does having a `None` here mean? Can't we guarantee that we always create these with a notifier?

crates/topos-tce-api/src/stream/mod.rs
```diff
@@ -31,6 +31,38 @@ pub use self::commands::StreamCommand;
...
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        self.inner.poll_recv(cx)
+    }
+}
+
+impl Drop for TransientStream {
+    fn drop(&mut self) {
+        if let Some(notifier) = self.notifier.take() {
```
dvdplm, line 60: Perhaps add a `tracing::trace!("Dropping TransientStream")` here?

dvdplm (commented):
One question I have is about how the system will behave when the channels are full for whatever reason. Would it be very messy to add a test for that?

crates/topos-tce-api/src/stream/mod.rs
Freyskeyd: It's a `oneshot::Sender` it moves when used, wrapping it in `Option` allows you to consume it one time exactly.

crates/topos-tce-api/src/runtime/mod.rs
Freyskeyd: A tuple ? for what exactly?

crates/topos-tce-api/src/runtime/mod.rs
Freyskeyd: The error on a `oneshot::Sender` is `T` so a `TransientStream`. It has no value to print it.

crates/topos-tce-api/src/graphql/query.rs
Freyskeyd: `SubscriptionRoot` is a struct used for setting up `GraphQL` schema.
`TransientStream` is described on the struct doc itself.

>  Q: this call looks like it doesn't really need to be async.
How would you do to make it non async then? The ownership of creating the different `Stream` is own by the `topos-tce-api::Runtime`.

> should we strive to make as much code as possible follow the oneshot channel pattern seen here, or only where it's strictly needed/useful?

The `oneshot` pattern is a request/response pattern, and it is the only (or at least the cleanest) to communicate asynchronously with a spawn task. It allows internal back pressure and prevents managing complex future pool and ids.

> Like, what is the point of sending the InternalRuntimeCommand::NewTransientStream and then wait for it, rather than do the work synchronously? (Honest question, not intended as a critique)

On one side we have the GraphQL API that run into its own task, on the other side we have the gRPC runtime that run also in its own task, and in the middle we have the topos API runtime which create the link between the rest of the internal topos component and those 2 frontend API. I'm curious to know how you would implement that without blocking everything.

Freyskeyd:
> One question I have is about how the system will behave when the channels are full for whatever reason. Would it be very messy to add a test for that?

Can you define the channels that you're talking about? We have many different kind

crates/topos-tce-api/src/graphql/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+                                }
+                              }
+                            }"
+            },
+        }))
+        .unwrap(),
+    )
+    .await
+    .unwrap();
+    // let response = client
```
gruberb, line 112: Remove comments

gruberb (changes requested):
Just a minor change: Removing the comments.

gruberb (approved):
LGTM 👍

crates/topos-tce-api/src/graphql/query.rs
dvdplm: What I had in mind to ask was essentially this code:

```rust
        use std::sync::mpsc::sync_channel;
        let (sender, receiver) = sync_channel(255);
        register.send(InternalRuntimeCommand::NewTransientStream { sender });
        let stream = receiver
            .recv()
            .map_err(|_| GraphQLServerError::InternalError("Bad"))?
            .map_err(|e| GraphQLServerError::TransientStream(e.to_string()))?;
        Ok(stream.map(|c| c.into()))
```

Why would the above not work? It would block (obviously), but we are unable to make progress on setting up the stream anyway so why not simply block until the other side has time to do the work? We `await` twice here, paying the overhead of pausing the task and yield, then awake and resume twice. Are we really creating so many new transient streams that it's worth paying the cost?

There is a cost to complex code and async rust **is** harder than sync rust. Harder to read, teach and debug.

I'm not saying this is bad code. I would like us to stop and think about applying async patterns sometimes. And be careful to apply it only when it's needed.

Freyskeyd: Async rust is pretty straightforward and documented, both in `Tokio` and standard library.
Making things blocking for no reason is a bad design I think, blocking stuff must be used only for critical things  or internal mutation, which is not the case here.

The `sync_channel` blocks the current thread, preventing any future to progress until the tasks finish. Do we want to block an entire thread for each API calls that could create a stream? I don't think so.

> I would like us to stop and think about applying async patterns sometimes. And be careful to apply it only when it's needed.

Don't worry :)
We know where we need to remove async stuff on the codebase. To stabilize the project we split the whole codebase into multiple chunks to have a better separation of concern between domain and logic. We are reducing those components now that we have a better view.

dvdplm: I'm glad you're finding it straight forward. :)
Many think otherwise, and regardless of how easy or hard it is for people, there is little doubt that debugging async code is (much) harder than sync code.

> Do we want to block an entire thread for each API calls that could create a stream? I don't think so.

You are correct of course. Did you know that I didn't realise this was being called once for each connection until I read this comment? I.e. that it's part of the setup for each graphql client. When I keep on harping on the need to document public methods this is what I mean: there will be readers of this code that are slow and inexperienced with the system and when we add a line of documentation, even if it's just stating the obvious, we end up saving readers like that some precious time.

crates/topos-tce-api/src/runtime/mod.rs
dvdplm: I was thinking: the `transient_stream` seems to be a single stream, not a collection, so why store the value in a `HashMap` as opposed to just having a `(Uuid, Sender<Certificate>)`. But maybe it should simply be renamed to `transient_streams` (plural)?

crates/topos-tce-api/src/graphql/query.rs
```diff
@@ -105,3 +110,52 @@ impl QueryRoot {
         Self::certificate_by_id(ctx, certificate_id).await
     }
 }
+
+pub struct SubscriptionRoot;
+
+impl SubscriptionRoot {
+    pub(crate) async fn new_transient_stream(
```
dvdplm, line 117:
```suggestion
    /// Set up a [`Stream`] of delivered [`Certificates`] on behalf of a new GraphQL client.
    pub(crate) async fn new_transient_stream(
```

crates/topos-tce-api/src/graphql/tests.rs
```diff
@@ -0,0 +1,121 @@
...
+use futures::{SinkExt, StreamExt};
+use rstest::rstest;
+use test_log::test;
+use tokio::sync::{mpsc, oneshot};
+use uuid::Uuid;
+
+#[rstest]
+#[test(tokio::test)]
+#[timeout(Duration::from_secs(2))]
+async fn requesting_transiant_stream_from_graphql() {
```
dvdplm, line 18:
```suggestion
async fn requesting_transient_stream_from_graphql() {
```

crates/topos-tce-api/src/runtime/mod.rs
dvdplm: I was thinking: the transient_stream seems to be a single stream, not a collection, so why store the value in a HashMap as opposed to just having a (Uuid, Sender<Certificate>). But maybe it should simply be renamed to transient_streams (plural)?

crates/topos-tce-api/src/graphql/query.rs
Freyskeyd: I mean, the `GraphQL` method that use this method is explaining a lot of things already.
The type `TransientStream` is also explain a lot of things about how it works.
I don't think we need to re-explain everything again for the third time.

crates/topos-tce-api/src/runtime/mod.rs
Freyskeyd: yes let's rename it to plural form

dvdplm (approved):

PR #330, opened 2023-10-06
feat: node flag for external edge

atanmarko:
# Description

For `topos node up` cli command, added flag `--external-edge-node` indicating that edge process should not be started.

Sequencer will connect to the edge node pointed by configuration

Fixes TP-741

## PR Checklist:

### DIFF
--- a/crates/topos/src/components/node/commands/up.rs
+++ b/crates/topos/src/components/node/commands/up.rs
@@ -12,4 +12,9 @@ pub struct Up {
     /// If omitted, the local FS secrets manager is used
     #[arg(long, env = "TOPOS_SECRETS_MANAGER")]
     pub secrets_config: Option<String>,
+
+    /// Do not run background edge process as part of node
+    /// Usable for cases where edge edpoint is available as infura (or similar cluod provider) endpoint
+    #[arg(long, env = "EXTERNAL_EDGE_NODE", action)]
+    pub external_edge_node: bool,
 }
--- a/crates/topos/src/components/node/commands/up.rs
+++ b/crates/topos/src/components/node/commands/up.rs
@@ -13,8 +13,8 @@ pub struct Up {
     #[arg(long, env = "TOPOS_SECRETS_MANAGER")]
     pub secrets_config: Option<String>,
 
-    /// Do not run background edge process as part of node
-    /// Usable for cases where edge edpoint is available as infura (or similar cluod provider) endpoint
-    #[arg(long, env = "EXTERNAL_EDGE_NODE", action)]
-    pub external_edge_node: bool,
+    /// Defines that an external edge node will be use, replacing the one normally run by the node.
+    /// Usable for cases where edge endpoint is available as infura (or similar cloud provider) endpoint
+    #[arg(long, env = "TOPOS_NO_EDGE_PROCESS", action)]
+    pub no_edge_process: bool,
 }
--- a/crates/topos/src/components/node/mod.rs
+++ b/crates/topos/src/components/node/mod.rs
@@ -138,24 +138,27 @@ pub(crate) async fn handle_command(
 
             let mut processes = FuturesUnordered::new();
 
-            // Edge
+            // Edge node
             if cmd.no_edge_process {
                 info!("Using external edge node, skip running of local edge instance...")
-            } else {
+            } else if let Some(edge_config) = config.edge {
                 let data_dir = node_path.clone();
                 info!(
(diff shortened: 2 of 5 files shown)


crates/topos/src/components/node/mod.rs
```diff
@@ -140,25 +139,45 @@ pub(crate) async fn handle_command(
...
+            if cmd.external_edge_node {
+                info!("Using external edge node, skip running of local edge instance...")
+            } else {
+                let data_dir = node_path.clone();
+                info!(
+                    "Spawning edge process with genesis file: {}, data directory: {}, additional \
+                     edge arguments: {:?}",
+                    genesis.path.display(),
+                    data_dir.display(),
+                    config.edge.as_ref().unwrap().args
```
dvdplm, line 151: No `unwrap` please.

crates/topos/src/components/node/commands/up.rs
```diff
@@ -12,4 +12,9 @@ pub struct Up {
     /// If omitted, the local FS secrets manager is used
     #[arg(long, env = "TOPOS_SECRETS_MANAGER")]
     pub secrets_config: Option<String>,
+
+    /// Do not run background edge process as part of node
+    /// Usable for cases where edge edpoint is available as infura (or similar cluod provider) endpoint
```
Freyskeyd, line 17:
```suggestion
    /// Usable for cases where edge endpoint is available as infura (or similar cloud provider) endpoint
```

crates/topos/src/components/node/commands/up.rs
```diff
@@ -12,4 +12,9 @@ pub struct Up {
     /// If omitted, the local FS secrets manager is used
     #[arg(long, env = "TOPOS_SECRETS_MANAGER")]
     pub secrets_config: Option<String>,
+
+    /// Do not run background edge process as part of node
```
Freyskeyd, line 16:
```suggestion
    /// Defines that an external edge node will be use, replacing the one normally run by the node.
```

crates/topos/src/components/node/commands/up.rs
```diff
@@ -12,4 +12,9 @@ pub struct Up {
...
+    /// Usable for cases where edge edpoint is available as infura (or similar cluod provider) endpoint
+    #[arg(long, env = "EXTERNAL_EDGE_NODE", action)]
+    pub external_edge_node: bool,
```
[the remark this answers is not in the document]
Freyskeyd: I think the goal of this flag is to prevent the edge process to be start, right?
Maybe we can define something like `no-edge-process` ?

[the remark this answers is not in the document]
atanmarko: Right, we start process by default, `no-edge-process` is non standard request so I would call it that way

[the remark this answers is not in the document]
atanmarko: @dvdplm We can have localhost node on the system but not started by topos cli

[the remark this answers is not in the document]
atanmarko: On the second thought, yes lets use `start-edge` - we do not have other double negatives in the arguments I believe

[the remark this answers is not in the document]
atanmarko: Update - clap does not naturally deals with default bool arguments - they are flags for it, so either it exists or does not exists. So I am going back to --no-edge-process flag

crates/topos/src/components/node/mod.rs
```diff
@@ -148,7 +148,7 @@ pub(crate) async fn handle_command(
                      edge arguments: {:?}",
                     genesis.path.display(),
                     data_dir.display(),
-                    config.edge.as_ref().unwrap().args
+                    config.edge.as_ref().expect("valid edge configuration").args
```
dvdplm, line 151: I don't see where the `edge` key is actually validated so that we can be sure that it's there at this point.

atanmarko: If there is no edge configuration it does not make sense to startup node

atanmarko: Updated: fit to exit gracefully if edge config is missing

dvdplm: Lovely, thank you for humoring my nitpicks. :)

crates/topos/src/components/node/commands/up.rs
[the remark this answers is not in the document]
dvdplm: Ok, fair enough.

dvdplm (approved):

PR #331, opened 2023-10-09
feat(p2p/grpc): creating gRPC behaviour

Freyskeyd:
https://img.shields.io/badge/1%20PR-grpc%20over%20p2p-blue 

# Description 
Next PR: #332 

This PR adds the gRPC behaviour in order to use gRPC client/server with p2p TCP connections.

The goal is to be able to rely on protobuf schema to dictate our network behaviours, allowing any other services that use gRPC to communicate with us transparently.
It also prevents us to generate and deal with complex custom-made Client/Server.
It also eases the `request/response` between two nodes and encapsulate the logic into dedicated structs

## PR Checklist:


crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+        };
+
+        let connection = Connection {
+            id: connection_id,
+            address,
+            request_id: None,
+            channel: None,
+        };
+
+        self.connected.entry(peer_id).or_default().push(connection);
```
dvdplm, line 244: Would it work to use `or_insert_with()` here too?

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+    /// Handle the [`ConnectionEstablished`] event comming from the [`Swarm`]
+    /// and try to open a gRPC channel using a [`ConnectionHandler`].
+    fn on_connection_established(
+        &mut self,
+        ConnectionEstablished {
+            peer_id,
+            connection_id,
+            endpoint,
+            failed_addresses,
+            other_established,
```
dvdplm, line 229: I'm not sure I understand what this value means. Other connections with this same peer?

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+    fn try_connect(&mut self, peer_id: &PeerId) {
+        if let Some(connections) = self.connected.get_mut(peer_id) {
+            let connection = connections.first_mut();
+            if let Some(connection) = connection {
+                if let Some(OutboundConnectionRequest {
+                    request_id,
+                    notifier,
+                }) = self.pending_outbound_connections.get(peer_id)
+                {
+                    info!("gRPC Outbound connection established with {peer_id}");
```
dvdplm, line 334: Maybe downgrade this to a `debug!`?

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+            request_id: None,
+            channel: None,
+        };
+
+        self.connected.entry(peer_id).or_default().push(connection);
+
+        // TODO refactor this to better handle connection
+        // If there is no current established connection it means that it's the
+        // first connection with that peer
+        if other_established == 0 {
```
dvdplm, line 249: I guess that what I don't understand is under which conditions we could get a `ConnectionEstablished` after we have established the first connection?

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+            connection_id,
+            endpoint,
+            failed_addresses,
+            other_established,
+        }: ConnectionEstablished,
+    ) {
+        let address = match endpoint {
+            ConnectedPoint::Dialer { address, .. } => Some(address.clone()),
+            ConnectedPoint::Listener { .. } => None,
+        };
```
dvdplm, line 235: SO if we have an address then we called them, otherwise they called us, is that correct? Do we store the address in our `Connection` so we can re-dial? Or do we need it for other reasons?

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+            self.try_connect(&peer_id);
+        }
+    }
+
+    /// Starts the gRPC service if not already started
+    fn on_new_listener(&mut self, listener_id: ListenerId) {
+        if let Some(service) = self.service.take() {
+            let (tx, rx) = mpsc::unbounded_channel();
+            self.inbound_stream = Some(tx);
+            // TODO: Switch to serve_with_incoming_shutdown at some point
```
dvdplm, line 259: Would be good to have a ticket for this.

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+            tokio::spawn(service.serve_with_incoming(proxy::GrpcProxy::new(rx)));
+            info!("New gRPC proxy start and listening on {listener_id:?}");
```
dvdplm, line 261:
```suggestion
            info!("New gRPC proxy started and listening on {listener_id:?}");
```

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+        } else {
+            warn!("Tried to instantiate gRPC proxy but service is missing");
```
dvdplm, line 263:
```suggestion
            warn!("Tried to instantiate a gRPC proxy on {listener_id:?} but the service is missing");
```

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+        &mut self,
+        ConnectionClosed {
+            peer_id,
+            connection_id,
+            endpoint,
+            handler,
+            remaining_established,
+        }: ConnectionClosed<<Self as NetworkBehaviour>::ConnectionHandler>,
+    ) {
+        warn!("Connection {connection_id} closed with peer {peer_id}");
```
dvdplm, line 278: This is normal behaviour right? If yes, then `debug!` is more appropriate.

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+    /// Handle the [`DialFailure`] event comming from the [`Swarm`]
+    fn on_dial_failure(
+        &mut self,
+        DialFailure {
+            peer_id,
+            error,
+            connection_id,
+        }: DialFailure,
+    ) {
+        println!("Dial failure {:?}", error);
```
dvdplm, line 296:
```suggestion
        warn!("Dial failure {:?}", error);
```

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+                    {
+                        self.pending_events.push_back(ToSwarm::GenerateEvent(
+                            Event::OutboundFailure {
+                                peer_id,
+                                request_id,
+                                error: OutboundError::DialFailure,
+                            },
+                        ));
+
+                        let _ = notifier.send(Err(OutboundError::DialFailure));
```
dvdplm, line 316: It looks like we're loose track of what the error causing the dial failure. I'm not what happens next here, but whatever code is handling `Event::OutboundFailure` and `OutboundError::DialFailure` will not have access to the error. Is that what we want?

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+        // Sending event to both `Swarm` and `ConnectionHandler`
+        if let Some(ev) = self.pending_events.pop_front() {
+            return Poll::Ready(ev);
+        }
+
+        // When channel has been negotiated by the [`ConnectionHandler`] we need
+        // to update the [`Connection`] with the channel.
+        match self.pending_negotiated_channels.poll_next_unpin(cx) {
+            Poll::Ready(Some((Ok(channel), request_id, peer_id))) => {
+                tracing::debug!("gRPC channel ready for {} {}", peer_id, request_id);
```
dvdplm, line 466:
```suggestion
                debug!("gRPC channel ready for {} {}", peer_id, request_id);
```

crates/topos-p2p/src/behaviour/grpc/stream.rs
```diff
@@ -0,0 +1,144 @@
...
+
+impl Future for NegotiatedGrpcOutboundStream {
+    type Output = Result<GrpcStream, BoxError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let fut = self.project();
+        let x = fut
+            .stream_rx
+            .lock()
+            .unwrap()
```
dvdplm, line 51: I know of the stdlib's mutexe poisoning stuff, but I think we should use `expect("why it's reasonable to make the assumptions we make")` as a pattern in all our code. It amounts to the same thing, just a wee bit clearer to the reader that we know what we're doing?

crates/topos-p2p/src/behaviour/grpc/event.rs
```diff
@@ -0,0 +1,37 @@
...
+    OutboundSuccess {
+        peer_id: PeerId,
+        request_id: RequestId,
+        channel: Channel,
+    },
+
+    // InboundNegotiatedStream {
+    //     request_id: RequestId,
+    //     stream: libp2p::Stream,
+    // },
```
dvdplm, line 23: Needed for later PRs or left-overs?

crates/topos-p2p/src/behaviour/grpc/event.rs
```diff
@@ -0,0 +1,37 @@
...
+    InboundNegotiatedConnection {
+        request_id: RequestId,
+        connection_id: ConnectionId,
+    },
+
+    // OutboundNegotiatedStream {
+    //     request_id: RequestId,
+    //     stream: libp2p::Stream,
+    // },
```
dvdplm, line 32: Needed for later PRs or left-overs?

crates/topos-p2p/src/behaviour/grpc/connection.rs
```diff
@@ -0,0 +1,90 @@
...
+
+/// Connection struct which represent a connection between two nodes
+/// It contains the connection id, the address of the node, the request id
+/// and the gRPC channel which is used to communicate with the node.
+#[derive(Debug)]
+pub(crate) struct Connection {
+    /// The connection id
+    pub(crate) id: ConnectionId,
+
+    /// The address of the node
```
dvdplm, line 21: THe address of the **remote** node right?

crates/topos-p2p/src/behaviour/grpc/connection.rs
```diff
@@ -0,0 +1,90 @@
...
+#[derive(Debug)]
+pub struct OutboundConnectedConnection {
+    #[allow(dead_code)]
+    pub(crate) request_id: RequestId,
+    // TODO: Remove unused when gRPC behaviour is activated
+    #[allow(unused)]
+    pub(crate) channel: tonic::transport::Channel,
+}
+
+/// Enum which represent the different state of an outbound connection
```
dvdplm, line 47:
```suggestion
/// Enum which represent the different states of an outbound connection
```

dvdplm (approved):
This is a big PR, good job! Can't say I understood the whole thing fully, but overall I think it's looking good. I'm approving this, but it would be good if you addressed the various nitpicks. The one true concern I have is the possibility of race conditions on `RequestId`.

crates/topos-p2p/Cargo.toml
```diff
@@ -15,15 +15,30 @@ prometheus-client.workspace = true
...
 hex.workspace = true
 unsigned-varint = { version = "0.7.1", features = ["futures", "asynchronous_codec"] }
 void = "1"
+tonic = {workspace = true, features = ["tls", "tls-roots"]}
+tower.workspace = true
 
+topos-api = {path = "../topos-api/", features = ["grpc"]}
 topos-metrics = { path = "../topos-metrics/" }
 topos-crypto = { path = "../topos-crypto" }
+smallvec = "1.11.1"
```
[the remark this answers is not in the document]
Freyskeyd: Tinyvec expects that the element type implement `Default` which is not the case for `Connection`

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+/// to a peer without having a gRPC service to expose.
+pub(crate) struct Behaviour {
+    /// The optional gRPC service to expose
+    service: Option<Router>,
+    /// The next request id to use
+    next_request_id: RequestId,
+    /// The next inbound request id to use
+    next_inbound_request_id: Arc<AtomicU64>,
+    /// The list of connected peers with the associated gRPC channel
+    connected: HashMap<PeerId, SmallVec<[Connection; 2]>>,
```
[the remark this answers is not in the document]
Freyskeyd: 2 connections because we could have one `inbound` and one `outound`.
6 `multiaddr` because of the different transports (dns/tcp/ws/relay...)

crates/topos-p2p/src/behaviour/grpc.rs
Freyskeyd: Yes, it's the number of existing established connection with this peer

crates/topos-p2p/src/behaviour/grpc.rs
Freyskeyd: The behaviour need to expose addresses in order to connect or reconnect

crates/topos-p2p/src/behaviour/grpc.rs
Freyskeyd: `or_insert_with` would need to provide an empty `vec` and then `push` to it?
It is more clear and recommanded to rely on `Default`

crates/topos-p2p/src/behaviour/grpc.rs
Freyskeyd: If a node exposes 10 addresses, you can have 10 different connections for the same peer. Or even multiple connections per transport in order to avoid HOL blocking.

crates/topos-p2p/src/behaviour/grpc.rs
```diff
@@ -0,0 +1,516 @@
...
+        } else {
+            debug!("Buffering sender as no available connection to peer {peer_id} yet");
+            self.open_connection(peer_id)
+        }
+    }
+
+    /// Return the next outbound request id
+    fn next_request_id(&mut self) -> RequestId {
+        let request_id = self.next_request_id;
+        self.next_request_id.0 += 1;
```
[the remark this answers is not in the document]
Freyskeyd: > Isn't this racy?

I don't think I understand correctly, can you rephrase?

crates/topos-p2p/src/behaviour/grpc.rs
Freyskeyd: We don't really care about the reason of this `DialFailure`, we only care about the `condition` failure.

crates/topos-p2p/src/behaviour/grpc/connection.rs
Freyskeyd: It's the address of the node for this connection

crates/topos-p2p/src/behaviour/grpc.rs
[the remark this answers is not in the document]
Freyskeyd: If it's about race condition, because we have mutable reference on `self` we can't have race condition.

crates/topos-p2p/src/behaviour/grpc.rs
dvdplm: 💡 aha!

crates/topos-p2p/src/behaviour/grpc.rs
dvdplm: 🤷‍♂️ Fair enough.

crates/topos-p2p/src/behaviour/grpc/connection.rs
dvdplm: Hmm, so it's the address of whichever node dialed the other?

crates/topos-p2p/src/behaviour/grpc/connection.rs
dvdplm:
```suggestion
/// Enum that represents the different states of an outbound connection
```

crates/topos-p2p/src/behaviour/grpc.rs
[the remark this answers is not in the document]
dvdplm: Maybe update the doc comment with this info?

But why is it a growable vector if we know it's always a fixed size?

crates/topos-p2p/src/behaviour/grpc.rs
dvdplm: No, you'd build the content right away so no push needed. `self.connected.entry(peer_id).or_insert_with(|| vec![connection])` works I think.

I don't think it matters, but as a reader, the `or_default()` forced me to go look up what the default is, which isn't relevant in this case.

Freyskeyd: yeah but the behaviour is different and doesn't cover what is expected. Look this playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=10a11a2cea226ce8a965d1e0a58ad9ae

crates/topos-p2p/src/behaviour/grpc.rs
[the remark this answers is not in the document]
Freyskeyd: We don't know it's always a fixed size. We know that it is common to have 2 connection but we may have more, that the point of using `SmallVec`.

crates/topos-p2p/src/behaviour/grpc/connection.rs
Freyskeyd: Yes the address of the node represented by the connection

crates/topos-p2p/src/tests/behaviour/grpc.rs
```diff
@@ -111,7 +111,7 @@ async fn opening_outbound_stream() {}
 async fn opening_outbound_stream_half_close() {}
 
 #[test(tokio::test)]
-#[ignore = "Need to find a way to properly close the connection after sending the query"]
+#[ignore = "TP-757: Need to find a way to properly close the connection after sending the query"]
```
dvdplm, line 114: ❤️

PR #332, opened 2023-10-09
feat(api/grpc): add synchronization proto

Freyskeyd:
https://img.shields.io/badge/2%20PR-grpc%20over%20p2p-blue 

# Description
Previous PR: #331 
Next PR: #333 

This PR adds the definition of the `gRPC` synchronization protocol, which contains various structs and also service definition.
This protocol will evolve as we implement and update the different kind of synchronization.

## PR Checklist:

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -39,7 +39,16 @@ jobs:
         run: cargo nextest run --workspace --exclude topos-sequencer-subnet-runtime
 
       - name: Run grcov
-        run: grcov . --binary-path target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore '../**' --ignore '/*' -o coverage.lcov
+        run: |
+          grcov . --binary-path target/debug/ -s . \
+            -t lcov \
+            --branch \
+            --ignore-not-existing \
+            --ignore 'crates/topos-api/src/grpc/generated/*' \
+            --ignore 'crates/topos-test-sdk/src/grpc/behaviour/*' \
+            --ignore '../**' \
+            --ignore '/*' \
+            -o coverage.lcov
 
       - name: Upload to codecov.io
         uses: codecov/codecov-action@v3
--- a/crates/topos-api/tests/tce_layer.rs
+++ b/crates/topos-api/tests/tce_layer.rs
@@ -1,17 +1,19 @@
 use async_stream::stream;
 use futures::{channel::oneshot, FutureExt};
 use futures::{Stream, StreamExt};
+use rstest::rstest;
 use std::collections::HashMap;
 use std::pin::Pin;
 use std::time::Duration;
 use test_log::test;
 use tokio::sync::mpsc;
+use tonic::transport::Endpoint;
 use tonic::{transport::Server, Request, Response, Status, Streaming};
-use topos_api::grpc::shared;
 use topos_api::grpc::shared::v1::checkpoints::TargetCheckpoint;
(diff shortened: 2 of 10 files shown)


crates/topos-api/proto/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@topos-protocol/topos-grpc-client-stub",
-  "version": "0.1.6",
+  "version": "0.1.7",
```
dvdplm, line 3: So this is a manual thing we have to do for each .proto update? Are we using semver here?

crates/topos-api/tests/tce_layer.rs
```diff
@@ -255,3 +258,11 @@ async fn create_tce_layer() {
...
     jh.await.unwrap();
 }
+
+#[rstest]
+#[test(tokio::test)]
+async fn create_grpc_client() {
+    let entrypoint = Endpoint::from_static("http://127.0.0.1:1340").connect_lazy();
+
+    let _client = SynchronizerServiceClient::init(entrypoint);
+}
```
dvdplm, line 268: Not sure this test adds much value?

.github/workflows/coverage.yml
```diff
@@ -39,7 +39,16 @@ jobs:
...
 
       - name: Run grcov
-        run: grcov . --binary-path target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore '../**' --ignore '/*' -o coverage.lcov
+        run: |
+          grcov . --binary-path target/debug/ -s . \
+            -t lcov \
+            --branch \
+            --ignore-not-existing \
+            --ignore 'crates/topos-api/src/grpc/generated/*' \
+            --ignore 'crates/topos-test-sdk/src/grpc/behaviour/*' \
```
dvdplm, line 48: Shouldn't all of `topos-test-sdk` be ignored? Do we care about coverage for test utilities?

dvdplm (approved):
LGTM modulo nitpicks

crates/topos-api/proto/package.json
Freyskeyd: Yes but we could remove the whole package if I understood correctly @sebastiendan ?

crates/topos-api/tests/tce_layer.rs
Freyskeyd: It tests that we can initialize a client with a channel using the custom-made trait. It also make the patch coverage happy 😃

.github/workflows/coverage.yml
Freyskeyd: Maybe, but as it is code that are used to build tests maybe we should keep them to spot unused or unreachable code blocks.

dvdplm: Fair enough.

crates/topos-api/tests/tce_layer.rs
dvdplm: > It also make the patch coverage happy 😃

This is exactly why I am skeptical of coverage measurements. It can lead to adding tests that we wouldn't otherwise, just to satisfy the "bureaucracy".

crates/topos-api/proto/package.json
sebastiendan: Yes, we can remove that npm package as a whole. It was initially used to build js stubs when the idea was to use grpc as the TCE public API, but then we realized grpc wasn't made for that purpose and we switched to graphql.

@Freyskeyd Is it planned in the Protocol team to remove the package, or should I open a PR?

gruberb (approved):
👍

PR #334, opened 2023-10-09
chore: update github actions

Freyskeyd:
# Description

Update github actions to allow PR of PR to run tests

## PR Checklist:

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -2,10 +2,14 @@ name: Coverage
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - "**"
     paths: crates/**
   pull_request:
-    branches: [main]
+    branches:
+      - main
+      - "**"
     paths: crates/**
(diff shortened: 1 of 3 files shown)


.github/workflows/coverage.yml
```diff
@@ -2,10 +2,14 @@ name: Coverage
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - "*/*"
```
hadjiszs, line 7: I think we basically want to capture every branches here no? In which case we want `*/**` or `**` I believe

Freyskeyd: I was thinking that we could only build branches following the conventional commit, but as you wish

hadjiszs: Yeah, sometimes we are naming branches without slash for very quick changes without passing by linear tickets, so I guess we just want to run the main jobs for any non-draft PR regardless of the branch name

hadjiszs (approved):

PR #337, opened 2023-10-11
feat: add dedicated websocket cli parameter

atanmarko:
# Description

Add sequencer command line argument `--subnet-websocket-endpoint` to explicitly provided subnet jsonrpc websocket endpoint. Usable when subnet endpoint could not be derived from ` http/https` endpoint in a standard way (what is done by default), e.g. when custom cloud websocket endpoints are used.

Fixes TP-749

## PR Checklist:

### DIFF
--- a/crates/topos/src/components/sequencer/commands/run.rs
+++ b/crates/topos/src/components/sequencer/commands/run.rs
@@ -9,21 +9,22 @@ pub struct Run {
     #[clap(long, env = "TOPOS_LOCAL_SUBNET_ID")]
     pub subnet_id: Option<String>,
 
-    // Subnet endpoint in the form [ip address]:[port]
-    // Topos sequencer expects both websocket and http protocol available
-    // on this subnet endpoint
+    /// Subnet endpoint in the form [ip address]:[port]
+    /// Topos sequencer expects both websocket and http protocol available
+    /// on this subnet endpoint. If optional `subnet_jsonrpc_ws` is not provided websocket endpoint
+    /// will be deduced from this parameter.
     #[clap(
         long,
         default_value = "127.0.0.1:8545",
-        env = "SUBNET_JSONRPC_ENDPOINT"
+        env = "TOPOS_SUBNET_JSONRPC_HTTP"
     )]
-    pub subnet_jsonrpc_endpoint: String,
+    pub subnet_jsonrpc_http: String,
 
-    // Optional explicit websocket endpoint for the subnet jsonrpc api. If this parameter is not provided,
-    // it will be derived from the `subnet_jsonrpc_endpoint`.
-    // Full uri value is expected, e.g. `wss://arbitrum.infura.com/v3/ws/mykey` or `ws://127.0.0.1/ws`
-    #[clap(long, env = "SUBNET_WEBSOCKET_ENDPOINT")]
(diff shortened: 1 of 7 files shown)


crates/topos/src/components/sequencer/commands/run.rs
```diff
@@ -19,6 +19,12 @@ pub struct Run {
     )]
     pub subnet_jsonrpc_endpoint: String,
 
+    // Optional explicit websocket endpoint for the subnet jsonrpc api. If this parameter is not provided,
+    // it will be derived from the `subnet_jsonrpc_endpoint`.
+    // Full uri value is expected, e.g. `wss://arbitrum.infura.com/v3/ws/mykey` or `ws://127.0.0.1/ws`
+    #[clap(long, env = "SUBNET_WEBSOCKET_ENDPOINT")]
+    pub subnet_websocket_endpoint: Option<String>,
```
hadjiszs, line 26: I'd mention about that on the description of `subnet_jsonrpc_endpoint` too

(also `///` for the rustdoc)

hadjiszs (commented):
I think the initial idea is more to make it straight and not flexible, simply `subnet_http_jsonrpc_endpoint` and `subnet_ws_jsonrpc_endpoint` given explicitly, without any attempt to infer anything

Here the current patch is nice because pretty backward compatible with what we have now, but the concern is that's the type of weird logic that are not nice when they adds up I think

dvdplm (changes requested):
I think we should stop inferring things automatically, like @hadjiszs suggests.

Also, wrt naming: JSON-RPC is the *protocol* and http/ws are the available *transports*. With that in mind I think `subnet_websocket_endpoint` is a poor name. I suggest we rename the options to `subnet_jsonrpc_http` and `subnet_jsonrpc_ws` and make both required (no `Option<String>`).

I am a bit concerned about our naming of env vars; I think they should be consistently prefixed with `TOPOS_`, but that's something for a different PR. :)

atanmarko:
> I think we should stop inferring things automatically, like @hadjiszs suggests.
> 
> Also, wrt naming: JSON-RPC is the _protocol_ and http/ws are the available _transports_. With that in mind I think `subnet_websocket_endpoint` is a poor name. I suggest we rename the options to `subnet_jsonrpc_http` and `subnet_jsonrpc_ws` and make both required (no `Option<String>`).
> 
> I am a bit concerned about our naming of env vars; I think they should be consistently prefixed with `TOPOS_`, but that's something for a different PR. :)

Right I am aware of that. I didn't want to break things with the tools team setup (to rename the subnet_jsonrpc_endpoint) and add topos to the env variable. Lets break everything then.

atanmarko:
@dvdplm @hadjiszs Updated param names to `subnet_jsonrpc_http` and `subnet_jsonrpc_ws`. I am strongly against making websocket argument required, as first time we needed it in a year was recently when Monir was dealing with the external subnets. Anyway lets leave that decision for some other PR.

dvdplm:
> I am strongly against making websocket argument required, as first time we needed it in a year was recently when Monir was dealing with the external subnets. Anyway lets leave that decision for some other PR.

Ok, maybe you're right. We'll discuss it elsewhere.  :)

PR #349, opened 2023-10-25
chore: fixing requirement for docker github workflow

Freyskeyd:
# Description

Fixing the `requirement` of the docker GitHub workflow, removing the `require: true` to allow the use of the default value.

## PR Checklist:


atanmarko (approved):

crates/topos-tce-storage/src/validator/tables.rs
```diff
@@ -78,8 +86,9 @@ impl ValidatorPerpetualTables {
             ColumnFamilyDescriptor::new(cfs::UNVERIFIED, default_options()),
         ];
 
-        let db = init_with_cfs(&path, default_options(), cfs)
-            .unwrap_or_else(|_| panic!("Cannot open DB at {:?}", path));
+        let db = init_with_cfs(&path, default_options(), cfs).unwrap_or_else(|e| {
+            panic!("Cannot open DB at {:?} => error {:?}", path, e);
```
dvdplm, line 90: Do you think we should `error!("Cannot open DB at {:?} => error {:?}", path, e); exit(1)` here instead?

dvdplm (approved):

crates/topos-tce-storage/src/validator/tables.rs
Freyskeyd: Yes but not in this PR

gruberb (approved):
Very nice 💅

PR #352, opened 2023-10-26
fix: cleanup cli commands

gruberb:
# Description

Cleaning up not used `CLI` flags. WIP, putting it up there early to discuss.

Fixes TP-715

## Breaking changes

* Will break the local `docker` testing setup due to the usage of `tce keys`. 

## PR Checklist:


crates/topos/tests/snapshots/node__help_display.snap
```diff
@@ -7,9 +7,10 @@ Utility to manage your nodes in the Topos network
 Usage: topos node [OPTIONS] [COMMAND]
 
 Commands:
-  up    Spawn your node!
-  init  Setup your node!
-  help  Print this message or the help of the given subcommand(s)
+  up      Spawn your node!
+  init    Setup your node!
```
dvdplm, line 11: Total nitpick, but can we drop the exclamation marks? I don't want my CLI tools to yell at me...

crates/topos/src/tracing.rs
```diff
@@ -50,18 +39,14 @@ fn build_resources(otlp_service_name: String) -> Vec<KeyValue> {
...
 }
 
-fn create_filter(verbose: u8) -> EnvFilter {
-    if verbose > 0 {
-        EnvFilter::try_new(format!("warn,topos={}", verbose_to_level(verbose).as_str())).unwrap()
-    } else {
-        EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("warn,topos=info"))
-    }
+#[allow(dead_code)]
+fn create_filter() -> EnvFilter {
```
dvdplm, line 43: What is this code for? Why keep it around?

crates/topos/src/config/mod.rs
```diff
@@ -52,12 +48,12 @@ pub(crate) trait Config: Serialize {
     fn load(home: &Path, command: Option<Self::Command>) -> Result<Self::Output, figment::Error> {
         let mut figment = Figment::new();
 
-        figment = Self::load_from_file(figment, home);
-
         if let Some(command) = command {
             figment = Self::load_from_command(figment, command);
         }
 
+        figment = Self::load_from_file(figment, home);
```
dvdplm, line 55: Doesn't this change the priority of config values? Seems like now config file settings will trump command line settings? That would be surprising behaviour to me.

crates/topos/src/components/node/mod.rs
```diff
@@ -16,22 +25,36 @@ use crate::{
...
+};
 use topos_wallet::SecretManager;
 
 pub(crate) mod commands;
-pub mod services;
+pub(crate) mod services;
+
+pub(crate) struct NodeService {
+    pub(crate) console_client: Arc<Mutex<ConsoleServiceClient<Channel>>>,
+    pub(crate) _api_client: Arc<Mutex<ApiServiceClient<Channel>>>,
```
dvdplm, line 38: What you mean to signal with the leading `_`?

crates/topos/src/components/node/mod.rs
```diff
@@ -146,6 +172,12 @@ pub(crate) async fn handle_command(
 
             let shutdown_token = CancellationToken::new();
             let shutdown_trigger = shutdown_token.clone();
+
+            // Setup instrumentation if both otlp agent and otlp service name
+            // are provided as arguments
+            // We may want to use instrumentation in e2e tests
```
dvdplm, line 178: I don't understand what this last line means, do you think you can rephrase it?

crates/topos/src/components/node/mod.rs
```diff
@@ -206,28 +238,67 @@ pub(crate) async fn handle_command(
...
                     processes.clear();
                 }
             };
 
             Ok(())
         }
+
+        Some(NodeCommands::Status(status)) => {
+            let mut node_service = NodeService::with_grpc_endpoint(&status.node_args.node);
+            let exit_code = i32::from(!(node_service.call(status).await?));
```
dvdplm, line 258: What are the possible exit codes here and what do they mean?

crates/topos/src/components/node/services/process.rs
```diff
@@ -0,0 +1,137 @@
...
+
+use crate::config::genesis::Genesis;
+
+#[derive(Error, Debug)]
+pub enum Errors {
+    #[error("Failure on the TCE")]
+    TceFailure,
+    #[error("Failure on the Sequencer")]
+    SequencerFailure,
+    #[error("Failure on the Edge")]
```
dvdplm, line 25: Maybe "THING failure" is easier to read than "Failure on THING"?

crates/topos/src/components/node/services/process.rs
```diff
@@ -0,0 +1,137 @@
...
+        tce_grpc_endpoint: config.tce_grpc_endpoint,
+        signing_key: keys.validator.clone().unwrap(),
+        verifier: 0,
+        start_block: config.start_block,
+    };
+
+    debug!("Sequencer args: {config:?}");
+    spawn(async move {
+        topos_sequencer::run(config, shutdown).await.map_err(|e| {
+            error!("Failure on the Sequencer: {e:?}");
```
dvdplm, line 74:
```suggestion
            error!("Sequencer failure: {e:?}");
```

crates/topos/src/components/node/services/process.rs
```diff
@@ -0,0 +1,137 @@
...
+        auth_key: keys.network.map(AuthKey::PrivateKey),
+        signing_key: keys.validator.map(AuthKey::PrivateKey),
+        tce_addr: format!("/ip4/{}", config.libp2p_api_addr.ip()),
+        tce_local_port: config.libp2p_api_addr.port(),
+        tce_params: ReliableBroadcastParams::new(genesis.validator_count()),
+        api_addr: config.grpc_api_addr,
+        graphql_api_addr: config.graphql_api_addr,
+        metrics_api_addr: config.metrics_api_addr,
+        storage: StorageConfiguration::RocksDB(Some(config.db_path)),
+        network_bootstrap_timeout: Duration::from_secs(180),
```
dvdplm, line 102: 180 sec is pretty long!

crates/topos/src/components/node/commands/up.rs
```diff
@@ -18,4 +18,12 @@ pub struct Up {
     /// Usable for cases where edge endpoint is available as infura (or similar cloud provider) endpoint
     #[arg(long, env = "TOPOS_NO_EDGE_PROCESS", action)]
     pub no_edge_process: bool,
+    /// Socket of the opentelemetry agent endpoint.
```
dvdplm, line 21: "Socket"? Like a unix socket path? Or a hostname:port combo?

dvdplm (approved):
lgtm

crates/topos/src/components/node/services/process.rs
gruberb: Ping @hadjiszs. Was there a specific reason for `180`?

crates/topos/src/components/node/mod.rs
gruberb: We didn't modify this, just moved it around. It's a fair question though. @Freyskeyd, do you know?

gruberb: Just FYI, we are going to sunset the way we request the status from a node, so this will change soon. But we can add a comment if you want.

.github/workflows/docker_utils.yml
```diff
@@ -19,10 +18,6 @@ on:
         required: false
         type: string
         default: stable
-      features:
```
Freyskeyd, line 22: if you remove this you need to remove/update the related steps for the tag generation

crates/topos/src/components/mod.rs
```diff
@@ -1,13 +1,3 @@
-pub(crate) mod doctor;
-#[cfg(feature = "network")]
-pub(crate) mod network;
-#[cfg(feature = "node")]
 pub(crate) mod node;
-#[cfg(feature = "sequencer")]
-pub(crate) mod sequencer;
-#[cfg(feature = "setup")]
+pub(crate) mod regtest;
```
Freyskeyd, line 2: `regtest` isn't clear for me :/ 
I don't understand what's being this

crates/topos/src/components/node/mod.rs
```diff
@@ -16,22 +24,36 @@ use crate::{
...
+            _api_client: setup_api_tce_grpc(endpoint),
+        }
+    }
+}
 
 pub(crate) async fn handle_command(
     NodeCommand {
         subcommands,
-        verbose,
+        verbose: _verbose,
```
Freyskeyd, line 52: replace by `NodeCommand { subcommands, home, edge_path, .. }`

crates/topos/src/components/node/mod.rs
```diff
@@ -16,22 +24,36 @@ use crate::{
...
+};
 use topos_wallet::SecretManager;
 
 pub(crate) mod commands;
-pub mod services;
+pub(crate) mod services;
+
+pub(crate) struct NodeService {
+    pub(crate) console_client: Arc<Mutex<ConsoleServiceClient<Channel>>>,
+    pub(crate) _api_client: Arc<Mutex<ApiServiceClient<Channel>>>,
```
Freyskeyd, line 37: It's not needed anymore I think

crates/topos/src/components/node/mod.rs
Freyskeyd: The return of service execution is a boolean so the status can be either `0` or `1`

crates/topos/src/components/node/services/process.rs
Freyskeyd: Maybe we should use something like: `Edge error: {0}`

crates/topos/src/tracing.rs
```diff
@@ -50,18 +39,14 @@ fn build_resources(otlp_service_name: String) -> Vec<KeyValue> {
...
+    EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("warn,topos=info"))
 }
 
 // Setup tracing
 // If otlp agent and otlp service name are provided, opentelemetry collection will be used
 pub(crate) fn setup_tracing(
-    verbose: u8,
```
Freyskeyd, line 64: why did you removed the `verbose` ?

crates/topos/src/components/node/services/process.rs
hadjiszs: Hm I'm not sure, if I recall correctly this timeout was set empirically because of the time it takes in some tests that involve docker and docker compose but I'm not entirely sure. @Freyskeyd probably knows more about it

We can try reducing it and see if it breaks any existing tests

crates/topos/src/components/mod.rs
gruberb: It's just naming? We can call it utils, or whatever! We have two commands related to the testnet and/or local setup! 

Other projects use the regtest name for this. But we can choose whatever!

Or what isn't clear?

crates/topos/src/tracing.rs
gruberb: I didn't see it being used in any commands we used or people use! 

We can keep it but I wanted to clear up things we don't need!

Freyskeyd: There is a `verbose` attribute on `NodeCommand` and it was present also on `TceCommand`.

This is a global attribute that are present for subcommand. It allows us to have logs when executing the command by passing a flag instead of an environment variable.

crates/topos/src/components/node/services/process.rs
dvdplm: Ok, let's do 90 and see what happens?

crates/topos/src/tracing.rs
dvdplm: I think a `verbose` argument is not so great, it's too coarse. What we really would want is a `--log` switch that takes the same syntax as `RUST_LOG`, allowing all of these to do the expected thing:

- `--log trace`: set `TRACE` level logging
- `--log tce=warn,sequencer=trace`: set `WARN` for the tce and `TRACE` for sequencer and leave `INFO` for all other components

Freyskeyd: The fact is that having this approach bound the exposed log syntax of the inner architecture of the repo.

I don't have strong opinions on that but we need to be careful about how we design that, I find the `RUST_LOG` syntax powerful but annoying at the same time.

crates/topos/src/config/sequencer.rs
```diff
@@ -1,6 +1,6 @@
 use std::path::Path;
 
-use crate::components::node::commands::Up;
+use crate::components::node::commands::Init;
```
atanmarko, line 3: Why do you need this import?

crates/topos/src/config/tce.rs
```diff
@@ -7,7 +7,7 @@ use figment::{
 };
 use serde::{Deserialize, Serialize};
 
-use crate::components::node::commands::Up;
+use crate::components::node::commands::Init;
```
atanmarko, line 10: Also this one and similar commands imports?

crates/topos/src/tracing.rs
gruberb: Fair points. I reverted to the previous state when it comes to `verbose`, and we can decide in another PR on how we want to adjust/proceed with that?

.github/workflows/docker_utils.yml
gruberb: You mean this can be removed?

https://github.com/topos-protocol/topos/pull/352/files#diff-d375c297193279683833c8534db2366059cefe310fb8a33c6bba4b765492c145L76

JDawg287:
The docker-compose setup is failing again after the latest changes.
```
$ docker logs boot
⚙️ Reading the configuration from /tmp/node_config/node/boot/config.toml
2023-11-14T19:09:40.166011Z  INFO topos::config::genesis: Reading subnet genesis file /tmp/node_config/subnet/topos/genesis.json
2023-11-14T19:09:40.180451Z  INFO topos::components::node: 🧢 New joiner: boot for the "topos" subnet as Validator
Error: SetGlobalDefaultError("a global default trace dispatcher has already been set")
```

crates/topos/src/components/regtest/commands.rs
```diff
@@ -0,0 +1,34 @@
+use clap::{Args, Subcommand};
+
+mod push_certificate;
+mod spam;
+
+pub(crate) use push_certificate::PushCertificate;
+pub(crate) use spam::Spam;
+
+/// Run test commands (e.g., pushing a certificate to a TCE node)
```
hadjiszs, line 9:
```suggestion
/// Run test commands (e.g., pushing a certificate to a TCE process)
```

crates/topos/src/components/regtest/commands/push_certificate.rs
```diff
@@ -3,6 +3,7 @@ use clap::Args;
 use crate::options::input_format::InputFormat;
 
 #[derive(Args, Debug)]
+#[command(about = "Push a certificate to a TCE node")]
```
hadjiszs, line 6:
```suggestion
#[command(about = "Push a certificate to a TCE process")]
```

nit on the terminology, just to not let users think that they can run one "TCE node" somewhere, but instead can reach the "TCE process" of one Topos subnet node

crates/topos/tests/config.rs
```diff
@@ -221,5 +218,7 @@ async fn test_command_init_precedence_cli_env() -> Result<(), Box<dyn std::error
     assert!(config_contents.contains("role = \"sequencer\""));
     assert!(config_contents.contains("subnet = \"topos-cli\""));
 
+    remove_dir_all("/tmp/topos/test_command_init_precedence_cli").unwrap();
```
hadjiszs, line 221: shouldn't it be `tmp_home_dir`?

atanmarko (approved):

hadjiszs (approved):

crates/topos/tests/config.rs
gruberb: No, we use the `tmp_home_dir` just for the polygon edge install in this case. I removed this hand-set folder, because if you re-run the tests locally, it fails because the folder is not automatically removed.

PR #363, opened 2023-10-31
feat: refactoring the sync task for state synchronisation between nodes

gruberb:
# Description

Refactor the asynchronous sync task for the node certificate storage.

Fixes TP-736

## PR Checklist:


crates/topos-tce-api/src/runtime/mod.rs
```diff
@@ -137,6 +137,10 @@ impl Runtime {
                 Some(command) = self.runtime_command_receiver.recv() => {
                     self.handle_runtime_command(command).await;
                 }
+
+                Some(result) = self.running_tasks.next() => {
+                    info!("Task result");
```
Freyskeyd, line 142: maybe put this in `debug` and log the `result` ?

crates/topos-tce-api/src/runtime/mod.rs
```diff
@@ -51,11 +45,17 @@ pub(crate) use self::commands::InternalRuntimeCommand;
...
+
 pub(crate) type Streams =
-    FuturesUnordered<Pin<Box<dyn future::Future<Output = Result<Uuid, StreamError>> + Send>>>;
+    FuturesUnordered<Pin<Box<dyn Future<Output = Result<Uuid, StreamError>> + Send>>>;
 
 pub struct Runtime {
-    pub(crate) sync_tasks: HashMap<Uuid, JoinHandle<()>>,
+    /// Map of sync tasks and their stream id, so we can cancel them when a new stream
+    /// with the same id is registered
+    pub(crate) tasks: HashMap<Uuid, oneshot::Sender<()>>,
```
Freyskeyd, line 56: Instead of `oneshot` maybe we can switch to `CancellationToken`

crates/topos-tce-api/src/runtime/mod.rs
```diff
@@ -320,93 +325,12 @@ impl Runtime {
...
-                                            target_subnet_id,
-                                            source_subnet_id,
-                                            position: 0,
-                                            certificate_id: None,
-                                        });
-                                    }
-                                }
-                            }
+                    let (exit_sender, mut task) =
+                        Task::new(stream_id, target_subnet_stream_positions, storage, notifier);
```
Freyskeyd, line 329: Can we rename this as `SynchronizationTask` as we have multiple `task` now

crates/topos-tce-api/src/runtime/task.rs
```diff
@@ -0,0 +1,173 @@
...
+    Running,
+    /// The sync task failed and reported an error
+    Error,
+    /// The sync task exited gracefully and is done pushing certificates to the stream
+    Done,
+    /// The sync task was cancelled externally
+    Cancelled,
+}
+
+pub(crate) struct Task {
```
Freyskeyd, line 34: Add some documentation about what this `task` is doing, maybe rename it to something more clear like `SynchronizationTask` or any other alternative if you have better one.

crates/topos-tce-api/src/runtime/task.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub(crate) status: TaskStatus,
```
Freyskeyd, line 35: Can you document the fields please?

crates/topos-tce-api/src/runtime/task.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub(crate) stream_id: Uuid,
+    pub(crate) target_subnet_stream_positions: TargetSubnetStreamPositions,
+    pub(crate) storage: StorageClient,
+    pub(crate) notifier: Sender<StreamCommand>,
+    pub(crate) exit_signal: tokio::sync::oneshot::Receiver<()>,
+}
+
+impl Task {
+    pub(crate) fn new(
```
Freyskeyd, line 44: Missing doc here

crates/topos-tce-api/src/runtime/task.rs
```diff
@@ -0,0 +1,173 @@
...
+        stream_id: Uuid,
+        target_subnet_stream_positions: TargetSubnetStreamPositions,
+        storage: StorageClient,
+        notifier: Sender<StreamCommand>,
+    ) -> (tokio::sync::oneshot::Sender<()>, Self) {
```
Freyskeyd, line 49: `new` function usually returns `Self`, if you pass the `exit_receiver` as an argument it could make the function following Rust guideline.

crates/topos-tce-api/src/runtime/task.rs
```diff
@@ -0,0 +1,173 @@
...
+    type Output = TaskStatus;
+
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + Send + 'static>>;
+
+    fn into_future(mut self) -> Self::IntoFuture {
+        Box::pin(async move {
+            info!("Sync task started for stream {}", self.stream_id);
+            let mut collector: Vec<(CertificateDelivered, FetchCertificatesPosition)> = Vec::new();
+
+            for (target_subnet_id, mut source) in &mut self.target_subnet_stream_positions {
```
Freyskeyd, line 75: Can you create a ticket to refactor this chunk of code, the storage doesn't need async anymore so we can remove a lot of `await` etc

Freyskeyd (changes requested):

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -58,6 +58,7 @@ impl SyncTask {
         notifier: Sender<StreamCommand>,
         cancel_token: CancellationToken,
     ) -> Self {
+        println!("CREATED A NEW TASK");
```
Freyskeyd, line 61:
```suggestion
```

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -76,6 +77,7 @@ impl IntoFuture for SyncTask {
 
     fn into_future(mut self) -> Self::IntoFuture {
         Box::pin(async move {
+            println!("START FUTURE");
```
Freyskeyd, line 80:
```suggestion
```

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -0,0 +1,179 @@
...
+    Running,
+    /// The sync task failed and reported an error
+    Error,
+    /// The sync task exited gracefully and is done pushing certificates to the stream
+    Done,
+    /// The sync task was cancelled externally
+    Cancelled,
+}
+
+pub(crate) struct SyncTask {
```
Freyskeyd, line 37: Can we have some doc on this struct?

Freyskeyd (approved):

crates/topos-tce-api/src/runtime/mod.rs
```diff
@@ -137,6 +137,10 @@ impl Runtime {
                 Some(command) = self.runtime_command_receiver.recv() => {
                     self.handle_runtime_command(command).await;
                 }
+
+                Some(result) = self.running_sync_tasks.next() => {
+                    info!("The task has been completed with result: {:?}", result);
```
dvdplm, line 142: It's not very clear to me what "the task" actually is, and why – as a user – I'd want to see the result logged at the `info` level. It would be better imo to let tasks log their final result (if any and if relevant to users) rather than do it here?

crates/topos-tce-api/src/runtime/mod.rs
```diff
@@ -293,8 +297,9 @@ impl Runtime {
             } => {
                 info!("Stream {stream_id} is registered as subscriber");
```
dvdplm, line 298: Is `info` the right level here? Feels like a `debug` to me.

crates/topos-tce-api/src/runtime/mod.rs
```diff
@@ -293,8 +297,9 @@ impl Runtime {
...
 
-                if let Some(task) = self.sync_tasks.get(&stream_id) {
-                    task.abort();
+                if let Some(cancel_token) = self.sync_tasks.remove(&stream_id) {
```
dvdplm, line 300: What happens if the `sender.send()` call on line 312 fails? Maybe nothing at all, but I just wanted to double check that we understand what that failure condition means and if we're at all worried, consider moving this code here furthe down?

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -0,0 +1,179 @@
...
+use topos_core::uci::SubnetId;
+use topos_tce_storage::{FetchCertificatesFilter, FetchCertificatesPosition, StorageClient};
+use tracing::{error, info};
+use uuid::Uuid;
+
+type TargetSubnetStreamPositions = HashMap<SubnetId, HashMap<SubnetId, TargetStreamPosition>>;
+pub(crate) type RunningTasks =
+    FuturesUnordered<Pin<Box<dyn Future<Output = SyncTaskStatus> + Send>>>;
+
+/// Status of a sync task
```
dvdplm, line 21:
```suggestion
/// Status of a [`SyncTask`]
```

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -0,0 +1,179 @@
...
+///
+/// When registering a stream, a sync task is started to fetch certificates from the storage
+/// and push them to the stream.
```
dvdplm, line 24: Maybe this is more appropriate as docs for `SyncTask` itself?

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -0,0 +1,179 @@
...
+#[derive(Debug)]
+pub(crate) enum SyncTaskStatus {
+    ///  The sync task is active and started running
+    Running,
+    /// The sync task failed and reported an error
+    Error,
+    /// The sync task exited gracefully and is done pushing certificates to the stream
+    Done,
+    /// The sync task was cancelled externally
```
dvdplm, line 33: For my education, can you elaborate on what this "external cancellation" is? The only place I've seen is when a sync task is replaced by another one (so that would be good to add here), but I'm not sure why/when this happens.

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -0,0 +1,179 @@
...
+    Cancelled,
+}
+
+pub(crate) struct SyncTask {
+    /// The status of the SyncTask. Can be used to check if the task is still running
+    pub(crate) status: SyncTaskStatus,
+    /// The stream with which the SyncTask is connected and pushes certificates to
+    pub(crate) stream_id: Uuid,
+    /// The positions of each subnet in the stream of where the stream is currently left of
```
dvdplm, line 42: "…left of"? There's something wrong with this phrase I think. Can you rephrase it?

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -0,0 +1,179 @@
...
+
+impl IntoFuture for SyncTask {
+    type Output = SyncTaskStatus;
+
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + Send + 'static>>;
+
+    fn into_future(mut self) -> Self::IntoFuture {
+        Box::pin(async move {
+            println!("START FUTURE");
+            info!("Sync task started for stream {}", self.stream_id);
```
dvdplm, line 81: Is `info` the right level here?

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -0,0 +1,179 @@
...
+                }
+
+                for TargetStreamPosition {
+                    target_subnet_id,
+                    source_subnet_id,
+                    position,
+                    ..
+                } in source.values_mut()
+                {
+                    if self.cancel_token.is_cancelled() {
```
dvdplm, line 118: Why check this again? After line 85 we know the task is not cancelled yeah?

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -0,0 +1,179 @@
...
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + Send + 'static>>;
+
+    fn into_future(mut self) -> Self::IntoFuture {
+        Box::pin(async move {
+            println!("START FUTURE");
+            info!("Sync task started for stream {}", self.stream_id);
+            let mut collector: Vec<(CertificateDelivered, FetchCertificatesPosition)> = Vec::new();
+
+            for (target_subnet_id, source) in &mut self.target_subnet_stream_positions {
+                if self.cancel_token.is_cancelled() {
```
dvdplm, line 85: Why do we check the cancellation status inside the loop? Do we need to check it for every iteration?

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -0,0 +1,179 @@
...
+                        })
+                        .await
+                    {
+                        collector.extend(certificates_with_positions)
+                    }
+                }
+            }
+
+            for (CertificateDelivered { certificate, .. }, position) in collector {
+                info!(
```
dvdplm, line 140: Maybe `debug`?

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -0,0 +1,179 @@
...
+                                position: *position,
+                                certificate_id: Some(certificate.id),
+                            }],
+                            certificate,
+                        })
+                        .await
+                    {
+                        error!("Error sending certificate to stream: {}", e);
+                        self.status = SyncTaskStatus::Error;
+                        return SyncTaskStatus::Error;
```
dvdplm, line 166: Can't we let `SyncTaskStatus::Error` carry the error so that consumers can use it directly?

crates/topos-tce-api/src/runtime/sync_task.rs
gruberb: There is an `async` call happening in each loop, so I wonder, what if this call takes longer and we want to cancel in between as well?

crates/topos-tce-api/src/runtime/sync_task.rs
gruberb: `async` call inside the loop, so I thought this might take longer, and we want to cancel immedietaly if the same `uuid` stream is registering again.

PR #364, opened 2023-11-01
feat: use protobuf messages for double echo

atanmarko:
# Description

For double echo protocol, use protobuf message definitions and serialize messages using using grpc serialization instead of bincode.

Fixes TP-108

## Additions and Changes
* Changed `NetworkMessage`
* Removed `TceCommands` from `topos-tce-transport `
* Double echo proto messages and conversions defined in `topos-tce-api`
* ValidatorId moved to `topos-crypto`

## PR Checklist:


crates/topos-tce/src/app_context/network.rs
```diff
@@ -20,69 +24,99 @@ impl AppContext {
...
-                                .await
+                            let certificate_id = certificate_id.try_into().map_err(|e| {
+                                error!("Invalid certificate id, could not send Echo message: {e}");
+                                e
+                            });
+                            let validator_id = validator_id.try_into().map_err(|e| {
+                                error!("Invalid validator id, could not send Echo message: {e}");
+                                e
+                            });
+                            if let (Ok(certificate_id), Ok(validator_id)) =
```
Freyskeyd, line 76: Same can we have a log to see errors?

Freyskeyd (approved):
Some little updates need to be made but LGTM overall

crates/topos-tce/src/app_context/network.rs
atanmarko: Errors are printed above while converting `certificate_id` and `validator_id`. I'll add one more summary error.

PR #366, opened 2023-11-07
fix: add kebab-case rewrite for init and up commands

gruberb:
# Description

Fixed an issue where CLI arguments from the `node init` command where not taken due to auto-rewriting of `kebab-case`.

### DIFF
--- a/crates/topos/src/components/node/commands/init.rs
+++ b/crates/topos/src/components/node/commands/init.rs
@@ -15,8 +15,8 @@ pub struct Init {
     pub role: NodeRole,
 
     /// Subnet of your node
-    #[arg(long, env = "TOPOS_NODE_SUBNET_ID", default_value = "topos")]
-    pub subnet_id: Option<String>,
+    #[arg(long, env = "TOPOS_NODE_SUBNET", default_value = "topos")]
+    pub subnet: Option<String>,
 
     /// The path to the SecretsManager config file. Used for Hashicorp Vault.
     /// If omitted, the local FS secrets manager is used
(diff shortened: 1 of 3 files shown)


crates/topos/src/components/node/commands/init.rs
```diff
@@ -15,7 +15,7 @@ pub struct Init {
 
     /// Subnet of your node
     #[arg(long, env = "TOPOS_NODE_SUBNET", default_value = "topos")]
```
atanmarko, line 17:
```suggestion
    #[arg(long, env = "TOPOS_NODE_SUBNET_ID", default_value = "topos")]
```

atanmarko (approved):

crates/topos/src/components/node/commands/init.rs
gruberb: Great catch, thanks!

crates/topos/src/components/node/commands/init.rs
```diff
@@ -14,8 +15,8 @@ pub struct Init {
     pub role: NodeRole,
 
     /// Subnet of your node
-    #[arg(long, env = "TOPOS_NODE_SUBNET", default_value = "topos")]
-    pub subnet: Option<String>,
+    #[arg(long, env = "TOPOS_NODE_SUBNET_ID", default_value = "topos")]
+    pub subnet_id: Option<String>,
```
hadjiszs, line 19: Not really the subnet-id though, here that's the subnet alias that points to the genesis.

It can be changed to subnet-id when it'll effectively take the subnet-id (the Sequencer public key for now, and eventually the ICE-FROST public key)

gruberb: Should we change it in the `base` config than as well? I was just confused as to why we pass the argument as `subnet` but in the config its `subnet_id`:

```
#[serde(default = "default_subnet")]
pub subnet_id: String,
```

PR #367, opened 2023-11-08
feat: topos node init test

atanmarko:
# Description

Test cli and env paremeters for `topos node init`

Fixes TP-765

## PR Checklist:

### DIFF
--- a/crates/topos/tests/config.rs
+++ b/crates/topos/tests/config.rs
@@ -164,6 +164,7 @@ async fn test_command_init_name_precedence() -> Result<(), Box<dyn std::error::E
     std::fs::remove_dir_all(node_init_home_env)?;
 
     // Test node init with both cli and env flags
+    // Cli arguments should take precedence over env variables
     let node_init_home_cli = "/tmp/topos/test_command_init_name_precedence_cli";
     let node_edge_path_cli = polygon_edge_path(node_init_home_cli).await;
     let node_init_name_cli = "TEST_NODE_CLI";
--- a/crates/topos/tests/config.rs
+++ b/crates/topos/tests/config.rs
@@ -127,9 +127,9 @@ async fn test_handle_command_init_with_custom_name() -> Result<(), Box<dyn std::
 /// Test node init arguments precedence
 /// •  CLI flag should overwrite ENV variable
 #[tokio::test]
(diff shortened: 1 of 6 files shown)


crates/topos/tests/config.rs
```diff
@@ -123,3 +123,90 @@ async fn test_handle_command_init_with_custom_name() -> Result<(), Box<dyn std::
...
+
+    // Test node init with both cli and env flags
+    // Cli arguments should take precedence over env variables
+    let node_init_home_cli = "/tmp/topos/test_command_init_precedence_cli";
+    let node_edge_path_cli = polygon_edge_path(node_init_home_cli).await;
+    let node_init_name_cli = "TEST_NODE_CLI";
+    let node_init_role_cli = "sequencer";
+    let node_init_subnet_cli = "topos-cli";
+
+    let mut cmd = Command::cargo_bin("topos")?;
```
gruberb, line 174: I would personally like to have each command as a separate test case. It's easier to follow if something breaks, and easier to know if we covered "all" cases.

gruberb (changes requested):

crates/topos/tests/config.rs
```diff
@@ -123,3 +123,90 @@ async fn test_handle_command_init_with_custom_name() -> Result<(), Box<dyn std::
...
+        .join("node")
+        .join(node_init_name_env)
+        .join("config.toml");
+    assert!(config_path.exists());
+    // Check if config file params are according to env params
+    let config_contents = std::fs::read_to_string(&config_path).unwrap();
+    assert!(config_contents.contains("name = \"TEST_NODE_ENV\""));
+    assert!(config_contents.contains("role = \"fullnode\""));
+    assert!(config_contents.contains("subnet = \"topos-env\""));
+    std::fs::remove_dir_all(node_init_home_env)?;
```
dvdplm, line 164: We should use [`tempfile`](https://crates.io/crates/tempfile) instead of creating and cleaning up temp files. E.g. if one of the `assert`s above here fail, would the file be cleaned up?

crates/topos/tests/config.rs
dvdplm: That would be my preference as well.

gruberb (approved):
LGTM 👍

PR #371, opened 2023-11-10
feat: use websocket stream block listener

atanmarko:
# Description

Change block retrieval from the subnet node. Use websocket stream block subscription instead of timer. 

Fixes TP-599

## Additions and Changes

Old mechanism to retrieve block by number is kept for initial block sync on sequencer startup. When sync is performed, blocks are further received though block subscription stream.

## PR Checklist:

### DIFF
--- a/crates/topos-sequencer-subnet-client/src/lib.rs
+++ b/crates/topos-sequencer-subnet-client/src/lib.rs
@@ -9,10 +9,10 @@ use ethers::{
     abi::Token,
     core::rand::thread_rng,
     middleware::SignerMiddleware,
-    providers::{Http, Provider, ProviderError, Ws, WsClientError},
+    providers::{Http, Provider, ProviderError, StreamExt, Ws, WsClientError},
     signers::{LocalWallet, Signer, WalletError},
 };
-use ethers_providers::Middleware;
+use ethers_providers::{Middleware, SubscriptionStream};
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 use std::time::Duration;
@@ -57,6 +57,8 @@ pub struct BlockInfo {
 pub enum Error {
     #[error("new finalized block not available")]
     BlockNotAvailable(u64),
+    #[error("next stream block not available")]
+    StreamBlockNotAvailable,
     #[error("invalid block number: {0}")]
     InvalidBlockNumber(u64),
     #[error("data not available")]
@@ -127,6 +129,15 @@ impl SubnetClientListener {
         Ok(SubnetClientListener { contract, provider })
     }
 
+    pub async fn new_block_subscription_stream(
+        &self,
+    ) -> Result<SubscriptionStream<Ws, ethers::types::Block<ethers::types::H256>>, Error> {
+        self.provider
+            .subscribe_blocks()
+            .await
+            .map_err(Error::EthersProviderError)
+    }
+
     /// Subscribe and listen to runtime finalized blocks
     pub async fn get_finalized_block(
         &mut self,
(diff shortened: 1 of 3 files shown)


crates/topos-sequencer-subnet-runtime/src/proxy.rs
```diff
@@ -352,6 +363,34 @@ impl SubnetRuntimeProxy {
...
+        // Update certificate block history
+        certification.append_blocks(vec![block_info]);
+
+        let new_certificates = match certification.generate_certificates().await {
+            Ok(certificates) => certificates,
+            Err(e) => {
+                error!("Unable to generate certificates: {e}");
+                return Err(e);
+            }
+        };
```
dvdplm, line 383:
```suggestion
let new_certificates = certification.generate_certificates().await?;
```
Maybe?

crates/topos-sequencer-subnet-runtime/src/proxy.rs
```diff
@@ -352,6 +363,34 @@ impl SubnetRuntimeProxy {
...
+
+        debug!("Generated new certificates {new_certificates:?}");
+
+        for cert in new_certificates {
+            Self::send_new_certificate(subnet_runtime_proxy.clone(), cert, block_number).await
+        }
```
dvdplm, line 389: Dumb q: what happens if we don't `await` here? We don't expect a return result here and `send_new_certificate` cannot fail (right?), so perhaps "fire and forget" could work?

crates/topos-sequencer-subnet-client/src/lib.rs
```diff
@@ -194,6 +205,41 @@ impl SubnetClientListener {
...
+        stream: &mut SubscriptionStream<'_, Ws, ethers::types::Block<ethers::types::H256>>,
+    ) -> Result<BlockInfo, Error> {
+        if let Some(block) = stream.next().await {
+            let block_number = block.number.ok_or(Error::BlockNumberNotAvailable)?;
+            let events = match get_block_events(&self.contract, block_number).await {
+                Ok(events) => events,
+                Err(Error::EventDecodingError(e)) => {
+                    // FIXME: Happens in block before subnet contract is deployed, seems like bug in ethers
+                    error!("Unable to parse events from block {}: {e}", block_number);
+                    Vec::new()
```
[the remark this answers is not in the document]
atanmarko: Yes we just keep reading. I believe this issue is not much relevant because it happens only before Topos smart contracts are deployed

crates/topos-sequencer-subnet-runtime/src/proxy.rs
atanmarko: It will block here only if mpsc channel for passing certificates is full, otherwise it will resolve immediately.

Other general point: I am not if favor of  fire and forget approach in sequencer design. We should not create new certificates if pipeline for processing certificates is broken, better stop earlier. Anyway creating certificates depend on new block (which happens every 2 seconds), so not being able to pass new certificate in that time-frame is really some erroneous situation.

dvdplm: >  I am not if favor of fire and forget approach in sequencer design.

I'm not a big fan either, but isn't that exactly what we're doing here? We send off the cert and then `await`, but we don't look at the outcome. If `send_new_certificates` resolves, but with an `Err`, we wouldn't know right?

Whenever we can, I think we should avoid relying on unwritten assumptions like for instance the 2 sec blocktime.

crates/topos-sequencer-subnet-client/src/lib.rs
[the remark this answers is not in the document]
dvdplm: Fair!

crates/topos-sequencer-subnet-runtime/src/proxy.rs
atanmarko: He we log the error earlier if channel fails and return. We can discuss if we should panic or something else

atanmarko: I will update this mechanism to return error instead of logging it

atanmarko: @dvdplm I'll handle update of certificate generation and dispatching in a separate PR, as there are more things to optimize and cleanup

dvdplm (approved):

PR #373, opened 2023-11-14
feat: test topos node up execution

atanmarko:
# Description

Added test execution `topos node up` with parameters from environment and config file.

Fixes TP-765

## PR Checklist:

### DIFF
--- a/crates/topos/tests/config.rs
+++ b/crates/topos/tests/config.rs
@@ -1,11 +1,11 @@
 use assert_cmd::prelude::*;
-use std::fs::remove_dir_all;
+use std::ffi::OsString;
 use std::path::PathBuf;
-use std::process::Command;
+use std::process::{Command, ExitStatus};
 use tempfile::tempdir;
 use topos::install_polygon_edge;
 
-async fn polygon_edge_path(path: &str) -> String {
+async fn setup_polygon_edge(path: &str) -> String {
     let installation_path = std::env::current_dir().unwrap().join(path);
     let binary_path = installation_path.join("polygon-edge");
 
@@ -25,10 +25,43 @@ async fn polygon_edge_path(path: &str) -> String {
     installation_path.to_str().unwrap().to_string()
 }
 
+async fn generate_polygon_edge_genesis_file(
+    polygon_edge_bin: &str,
+    home_path: &str,
+    node_name: &str,
+    subnet: &str,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let genesis_folder_path: PathBuf = PathBuf::from(format!("{}/subnet/{}", home_path, subnet));
+    if !genesis_folder_path.exists() {
+        std::fs::create_dir_all(genesis_folder_path.clone())
+            .expect("Cannot create subnet genesis file folder");
+    }
+    let genesis_file_path = format!("{}/genesis.json", genesis_folder_path.display());
+    println!("Polygon edge path: {}", polygon_edge_bin);
+    let mut cmd = Command::new(polygon_edge_bin);
+    let val_prefix_path = format!("{}/node/{}/", home_path, node_name);
+    cmd.arg("genesis")
+        .arg("--dir")
+        .arg(&genesis_file_path)
+        .arg("--consensus")
+        .arg("ibft")
+        .arg("--ibft-validators-prefix-path")
+        .arg(val_prefix_path)
+        .arg("--bootnode") /* set dummy bootnode, we will not run edge to produce blocks */
+        .arg("/ip4/127.0.0.1/tcp/8545/p2p/16Uiu2HAmNYneHCbJ1Ntz1ojvTdiNGCMGWNT5MGMH28AzKNV66Paa");
+    let output = cmd.assert().success();
+    let output: &str = std::str::from_utf8(&output.get_output().stdout)?;
+    assert!(output.contains(&format!(
+        "Genesis written to {}",
+        genesis_folder_path.display()
+    )));
+    Ok(())
+}
+
 #[tokio::test]
 async fn handle_command_init() -> Result<(), Box<dyn std::error::Error>> {
     let tmp_home_dir = tempdir()?;
-    let path = polygon_edge_path(tmp_home_dir.path().to_str().unwrap()).await;
+    let path = setup_polygon_edge(tmp_home_dir.path().to_str().unwrap()).await;
 
     let mut cmd = Command::cargo_bin("topos")?;
     cmd.arg("node")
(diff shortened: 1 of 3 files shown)


crates/topos/tests/config.rs
```diff
@@ -24,6 +25,39 @@ async fn polygon_edge_path(path: &str) -> String {
     installation_path.to_str().unwrap().to_string()
 }
 
+async fn generate_polygon_edge_genesis_file(
```
gruberb, line 28: Not sure if we should just generate the `genesis.json` once and then copy it to the `temp_dir`?

crates/topos/tests/config.rs
```diff
@@ -223,3 +257,96 @@ async fn test_command_init_precedence_cli_env() -> Result<(), Box<dyn std::error
...
+        .env("TOPOS_HOME", node_up_home_env)
+        .env("TOPOS_NODE_NAME", node_up_name_env)
+        .env("TOPOS_NODE_ROLE", node_up_role_env)
+        .env("TOPOS_NODE_SUBNET", node_up_subnet_env)
+        .arg("init");
+
+    let output = cmd.assert().success();
+    let result: &str = std::str::from_utf8(&output.get_output().stdout)?;
+
+    // Run node init with cli flags
```
gruberb, line 285: Little nit pick. Here we `assert`, we run the command above.

crates/topos/tests/config.rs
```diff
@@ -223,3 +257,96 @@ async fn test_command_init_precedence_cli_env() -> Result<(), Box<dyn std::error
...
+    let mut child = subprocess::Popen::create(
+        &["cargo", "run", "--", "node", "up"],
+        subprocess::PopenConfig {
+            stdout: subprocess::Redirection::Pipe,
+            env: Some(environment),
+            ..Default::default()
+        },
+    )?;
+
+    // Use big timeout because `cargo run` sometimes builds before running
```
gruberb, line 330: This makes me a bit worried:
* This adds a significant time to our testing suite
* Makes the test a bit flaky because of `cargo run`

gruberb (commented):
I like the approach. I am just not sure to what extend we want to expand our test run time for this. I would personally rather test the error cases, and leave the `node up` test for larger `e2e` tests. 

We also don't test here if the right components are being triggered based on the config file. So I am not 100% sure. Maybe others can weigh in as well.

crates/topos/tests/config.rs
```diff
@@ -223,3 +257,96 @@ async fn test_command_init_precedence_cli_env() -> Result<(), Box<dyn std::error
...
+        (
+            "TOPOS_POLYGON_EDGE_BIN_PATH".into(),
+            (&node_edge_path_env).into(),
+        ),
+        ("TOPOS_HOME".into(), (&node_up_home_env).into()),
+        ("TOPOS_NODE_NAME".into(), (&node_up_name_env).into()),
+        ("RUST_LOG".into(), ("info").into()),
+    ]);
+
+    let mut child = subprocess::Popen::create(
```
Freyskeyd, line 321: Why not using [tokio::process::Command](https://docs.rs/tokio/latest/tokio/process/struct.Command.html) to be able to use the `timeout` directly on the command and avoid adding a new deps?

Something like that should works:

```rust
let std_cmd = Command::cargo_bin("topos").env(...);
let mut cmd = tokio::process::Command::from(std_cmd).spawn().unwrap();
timeout(Duration::from_secs(X), cmd.wait()).await.expect("Timeout on up command test");

// Check something with assert_cmd
cmd.as_std().assert().success();
...
```

crates/topos/tests/config.rs
atanmarko: I like every test to be independent. They are executed in random order.

atanmarko:
> I like the approach. I am just not sure to what extend we want to expand our test run time for this. I would personally rather test the error cases, and leave the `node up` test for larger `e2e` tests.
> 
> We also don't test here if the right components are being triggered based on the config file. > So I am not 100% sure. Maybe others can weigh in as well.

I think having basic `topos node up` test like this can save us some time with regressions while developing, especially changes related to topos cli. I don't see how it hurts, and why should we rely only on distant and ambiguous e2e tests for this basic use-case.

crates/topos/tests/config.rs
atanmarko: Subprocess has much better handling of the terminal output - you are able to capture it while subprocess works which is not able with Command - that is the reason I introduced subprocess. In the end output is not used because of the logs not working, so returning back to command is indeed possible.

crates/topos/tests/config.rs
atanmarko: I don't know - we have other tests with big timeout. @hadjiszs what do you think?

crates/topos/tests/config.rs
```diff
@@ -24,6 +25,39 @@ async fn polygon_edge_path(path: &str) -> String {
...
+    subnet: &str,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let genesis_folder_path: PathBuf = PathBuf::from(format!("{}/subnet/{}", home_path, subnet));
+    if !genesis_folder_path.exists() {
+        std::fs::create_dir_all(genesis_folder_path.clone())
+            .expect("Cannot create subnet genesis file folder");
+    }
+    let genesis_file_path = format!("{}/genesis.json", genesis_folder_path.display());
+    println!("Polygon edge path: {}", polygon_edge_bin);
+    let mut cmd = Command::new(polygon_edge_bin);
```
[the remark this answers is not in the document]
atanmarko: It is downloaded in the temp folder at the beginning of the test in `polygon_edge_path`. I'll rename this function to be more clear.

crates/topos/tests/config.rs
```diff
@@ -24,6 +25,39 @@ async fn polygon_edge_path(path: &str) -> String {
...
+        .arg("--ibft-validators-prefix-path")
+        .arg(val_prefix_path)
+        .arg("--bootnode") /* set dummy bootnode, we will not run edge to produce blocks */
+        .arg("/ip4/127.0.0.1/tcp/8545/p2p/16Uiu2HAmNYneHCbJ1Ntz1ojvTdiNGCMGWNT5MGMH28AzKNV66Paa");
+    let output = cmd.assert().success();
+    let output: &str = std::str::from_utf8(&output.get_output().stdout)?;
+    assert!(output.contains(&format!(
+        "Genesis written to {}",
+        genesis_folder_path.display()
+    )));
```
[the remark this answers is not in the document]
atanmarko: Tried it - I need to add `assert_cmd` and `predicates` as dependencies so I gave up.

gruberb (approved):
LGTM 👍

crates/topos/tests/config.rs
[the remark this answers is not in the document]
Freyskeyd: `assert_cmd` is already [part](https://github.com/topos-protocol/topos/blob/main/crates/topos/Cargo.toml#L65) of the deps and you're using it at line 52.

```rust
Command::cargo_bin("topos")
    ...
    .assert()
    .success()
    .stdout(predicate::str::contains(&format!(
        "Genesis written to {}",
        genesis_folder_path.display()
    )));

```

PR #374, opened 2023-11-17
fix: doc building actions

Freyskeyd:
# Description

Fixing doc warning and adding test actions

## PR Checklist:

### DIFF
--- a/crates/topos-tce-storage/src/index/mod.rs
+++ b/crates/topos-tce-storage/src/index/mod.rs
@@ -14,13 +14,10 @@ use crate::{
         db::{default_options, init_with_cfs},
         db_column::DBColumn,
     },
-    types::{CertificateSequenceNumber, TargetSourceListColumn, TargetStreamsColumn},
+    types::{TargetSourceListColumn, TargetStreamsColumn},
 };
 
-#[allow(unused)]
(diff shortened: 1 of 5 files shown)


crates/topos-tce-storage/src/index/mod.rs
```diff
@@ -17,8 +17,8 @@ use crate::{
     types::{CertificateSequenceNumber, TargetSourceListColumn, TargetStreamsColumn},
 };
 
-#[allow(unused)]
-pub(crate) struct IndexStore {
+pub struct IndexStore {
+    #[allow(unused)]
     certificate_order: DBColumn<CertificateSequenceNumber, CertificateId>,
```
atanmarko, line 22: Why do we need this field?

atanmarko (approved):

crates/topos-tce-storage/src/index/mod.rs
Freyskeyd: For now we don't need it, it was planned to use it for the API queries

atanmarko: Why don't you add it when we actually need it?

dvdplm (approved):

PR #379, opened 2023-11-20
chore: run tce only if there is tce config

atanmarko:
# Description

Check of existence  of  tce component configuration before running tce process.

## PR Checklist:

### DIFF
--- a/crates/topos/src/components/node/mod.rs
+++ b/crates/topos/src/components/node/mod.rs
@@ -221,13 +221,15 @@ pub(crate) async fn handle_command(
 
             // TCE
             if config.base.subnet == "topos" {
-                info!("Running topos TCE service...",);
-                processes.push(services::process::spawn_tce_process(
-                    config.tce.clone().unwrap(),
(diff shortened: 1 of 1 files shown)


crates/topos/src/components/node/mod.rs
```diff
@@ -221,13 +221,15 @@ pub(crate) async fn handle_command(
...
             // TCE
             if config.base.subnet == "topos" {
-                info!("Running topos TCE service...",);
-                processes.push(services::process::spawn_tce_process(
-                    config.tce.clone().unwrap(),
-                    keys,
-                    genesis,
-                    (shutdown_token.clone(), shutdown_sender.clone()),
-                ));
+                if let Some(tce_config) = config.tce {
```
gruberb, line 224: We can remove the `unwrap` and fail more gracefully, but we still need to fail.

gruberb (changes requested):

atanmarko:
Cancelled per discussion

PR #383, opened 2023-11-23
feat: add batch message and update double echo

atanmarko:
# Description

Cleanup double echo protoc from bincode serialization. 
Refactor a bit message pasing

Fixes TP-788

## PR Checklist:


crates/topos-api/proto/topos/tce/v1/double_echo.proto
```diff
@@ -30,3 +30,7 @@ message DoubleEchoRequest {
     Ready ready = 3;
   }
 }
+
+message Batch {
```
Freyskeyd, line 34: I'm wondering if creating a new proto file just for GossipSub related message could be beneficial, `Batch` isn't really part of the `double_echo` it is only an Envelope for the GossipSub protocol.

We could have something like:

```protobuf
file: `proto/topos/tce/v1/gossipsub.proto`

package topos.tce.v1;

message Batch {
	repeated bytes messages = 1;
}
```

crates/topos-p2p/src/behaviour/gossip.rs
```diff
@@ -10,37 +10,35 @@ use libp2p::{
...
-    pub(crate) data: Vec<Vec<u8>>,
-}
-
 pub struct Behaviour {
     batch_size: usize,
     gossipsub: gossipsub::Behaviour,
-    echo_queue: VecDeque<Vec<u8>>,
-    ready_queue: VecDeque<Vec<u8>>,
+    echo_queue: VecDeque<DoubleEchoRequest>,
+    ready_queue: VecDeque<DoubleEchoRequest>,
```
Freyskeyd, line 24: Do you think that we can remove those two and switch to a `HashMap<&'static str, VecDeque<Vec<u8>>>`  so we can remove the code duplication in publish/poll methods and also remove the link between the Gossip behaviour and the double echo?

crates/topos-p2p/src/behaviour/gossip.rs
```diff
@@ -10,37 +10,35 @@ use libp2p::{
...
     tick: tokio::time::Interval,
     cache: HashSet<MessageId>,
 }
 
 impl Behaviour {
-    pub fn publish(&mut self, topic: &'static str, data: Vec<u8>) -> Result<usize, &'static str> {
+    pub fn publish(
+        &mut self,
+        topic: &'static str,
+        message: DoubleEchoRequest,
```
Freyskeyd, line 33: It would be better to stick to `Vec<u8>` for flexibility

crates/topos-p2p/src/behaviour/gossip.rs
```diff
@@ -158,49 +156,42 @@ impl NetworkBehaviour for Behaviour {
...
+                    messages: Vec::new(),
+                };
                 for _ in 0..self.batch_size {
-                    if let Some(data) = self.echo_queue.pop_front() {
-                        echos.data.push(data);
+                    if let Some(request) = self.echo_queue.pop_front() {
+                        echos.messages.push(request);
                     } else {
                         break;
                     }
```
dvdplm, line 167: Would this work here?

```rust
let echoes = Batch{messages: self.echo_queue.drain(0..self.batch_size).collect()};
```

crates/topos-p2p/src/behaviour/gossip.rs
```diff
@@ -158,49 +156,42 @@ impl NetworkBehaviour for Behaviour {
...
+                    messages: Vec::new(),
+                };
                 for _ in 0..self.batch_size {
                     if let Some(data) = self.ready_queue.pop_front() {
-                        readies.data.push(data);
+                        readies.messages.push(data);
                     } else {
                         break;
                     }
                 }
```
dvdplm, line 189: Same here, looks like `drain` could help make this shorter.

crates/topos-p2p/src/behaviour/gossip.rs
```diff
@@ -158,49 +156,42 @@ impl NetworkBehaviour for Behaviour {
...
-                if let Ok(msg) = bincode::serialize::<Batch>(&readies) {
-                    P2P_GOSSIP_BATCH_SIZE.observe(readies.data.len() as f64);
-                    _ = self.gossipsub.publish(IdentTopic::new(TOPOS_READY), msg);
-                } else {
-                    P2P_MESSAGE_SERIALIZE_FAILURE_TOTAL
-                        .with_label_values(&["ready"])
-                        .inc();
-                }
+                debug!("Publishing {} readies", readies.messages.len());
+                let msg = readies.encode_to_vec();
```
dvdplm, line 192: This allocates again, so we're doing this three times I think? Once to put the message on the `VecDeque` queue, then again to build a `Vec` for the `Batch` and here one last time. Is there a way we can rethink this to avoid the overhead?

crates/topos-api/proto/topos/tce/v1/double_echo.proto
atanmarko: @Freyskeyd Problem I see with Batch having `repeated bytes messages` is how to de-serialize them. 

If we have `Batch` having other proto structure as its payload, we encode them at one point. If we first encode unknown number of multiple double echo messages to byte array, and then encode Batch, when we decode Batch on the other size we don't know how to decode `messages` and what is the payload meaning

atanmarko: @Freyskeyd Still it can not be just byte array. Because from one byte array, you don't know how big is the vector of byte array messages

crates/topos-p2p/src/behaviour/gossip.rs
dvdplm: I didn't mean to block this PR here, but perhaps there's some pattern to be figured out here that we can apply more broadly? 

/cc @Freyskeyd Any thoughts?

atanmarko: Well, simply protobuf Rust structure is different from serialized binary message of that sturcture. We can maybe optimize passing of protobuf structure prior to serialization

crates/topos-p2p/src/behaviour/gossip.rs
```diff
@@ -10,37 +11,32 @@ use libp2p::{
...
-    pub fn publish(&mut self, topic: &'static str, data: Vec<u8>) -> Result<usize, &'static str> {
+    pub fn publish(
+        &mut self,
+        topic: &'static str,
+        message: Vec<u8>,
+    ) -> Result<usize, &'static str> {
         match topic {
             TOPOS_GOSSIP => {
-                _ = self.gossipsub.publish(IdentTopic::new(topic), data);
+                _ = self.gossipsub.publish(IdentTopic::new(topic), message);
```
Freyskeyd, line 37: I'm wondering if we could not just `Batch` the `gossip` too 👀  what do you think?
Let's create a ticket for that as it is not really linked to this issue.

Freyskeyd (approved):

crates/topos-p2p/src/behaviour/gossip.rs
atanmarko: @dvdplm We are just passing message data around, we do not clone it. For the batch we temporary put it to the hash map queue and move it out from there. I don' think we can optimize allocation more than that

crates/topos-p2p/src/behaviour/gossip.rs
dvdplm: So this didn't work in the end?

atanmarko: https://github.com/topos-protocol/topos/pull/383/commits/e6a292ace6fea11b90d865ee1d2f28c4dd033eb3

PR #385, opened 2023-11-23
fix: update default values for api endpoints

gruberb:
# Description

Fix `0.0.0.1` to `0.0.0.0` 

## PR Checklist:

### DIFF
--- a/crates/topos/src/config/tce.rs
+++ b/crates/topos/src/config/tce.rs
@@ -52,28 +52,28 @@ fn default_db_path() -> PathBuf {
 
 const fn default_libp2p_api_addr() -> SocketAddr {
     SocketAddr::V4(std::net::SocketAddrV4::new(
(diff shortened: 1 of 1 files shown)


crates/topos/src/config/tce.rs
```diff
@@ -52,28 +52,28 @@ fn default_db_path() -> PathBuf {
 
 const fn default_libp2p_api_addr() -> SocketAddr {
     SocketAddr::V4(std::net::SocketAddrV4::new(
-        std::net::Ipv4Addr::new(0, 0, 0, 1),
+        std::net::Ipv4Addr::new(0, 0, 0, 0),
```
dvdplm, line 55: Can't these be `const` these days?

Freyskeyd: Yes it does. I also think that we need to merge this PR as soon as possible to have this kind of update in a unitary fashion

Freyskeyd (approved):

PR #399, opened 2023-12-06
chore: refactor push certificate tests and cleanup regtest

atanmarko:
# Description

Remove `regtest push-certificate` subcommand. Run standalone Rust test for related functionality.

Fixes TP-808

## PR Checklist:


crates/topos/tests/cert_delivery.rs
```diff
@@ -309,3 +313,209 @@ async fn cert_delivery() {
...
+        let response = client
+            .console_grpc_client
+            .status(StatusRequest {})
+            .await
+            .expect("Can't get status");
+
+        status.push(response.into_inner().has_active_sample);
+    }
+
+    assert!(status.iter().all(|s| *s));
```
hadjiszs, line 475:
```suggestion
    for (_peer_id, client) in peers_context.iter_mut() {
        let response = client
            .console_grpc_client
            .status(StatusRequest {})
            .await
            .expect("Can't get status");

        assert!(response.into_inner().has_active_sample));
    }
```

crates/topos/tests/cert_delivery.rs
```diff
@@ -309,3 +313,209 @@ async fn cert_delivery() {
         panic!("Timeout waiting for command");
     }
 }
+
+// Picks a random peer and sends it a certificate. All other peers listen for broadcast certs.
+// Three possible outcomes:
+// 1. No errors, returns Ok
+// 2. There were errors, returns a list of all errors encountered
+// 3. timeout
+async fn check_certificate_delivery(
```
hadjiszs, line 322: I know that this test isn't new and just get moved, but I'm not sure what is supposed to be the difference between the goal of this test and the `cert_delivery` test above in this same file. cc @Freyskeyd

crates/topos/tests/cert_delivery.rs
```diff
@@ -309,3 +313,209 @@ async fn cert_delivery() {
...
+    .map_err(|error| {
+        warn!("Timeout reached: {:?}", error);
+        error
+    })
+}
+
+async fn run_check_certificate_delivery(
+    number_of_nodes: usize,
+    timeout_broadcast: u64,
+    timeout: u64,
```
hadjiszs, line 459: Prefer using the Duration type directly from the 1st call

Also, not sure we need to deal with explicit sleeps, cant we just have one high level timeout through the `#[timeout`?

crates/topos/tests/cert_delivery.rs
```diff
@@ -309,3 +313,209 @@ async fn cert_delivery() {
...
+async fn push_and_deliver_cert_5() -> Result<(), Box<dyn std::error::Error>> {
+    run_check_certificate_delivery(5, 5, 20).await
+}
+
+#[rstest]
+#[test(tokio::test)]
+#[timeout(Duration::from_secs(20))]
+async fn push_and_deliver_cert_9() -> Result<(), Box<dyn std::error::Error>> {
+    run_check_certificate_delivery(9, 5, 20).await
+}
```
hadjiszs, line 521: I think you want to use the rstest feature for parameterized test cases with `#[case(args)]` (see [here](https://docs.rs/rstest/latest/rstest/attr.rstest.html#test-parametrized-cases)), easier to read+maintain 

You can see an example in our codebase as well, in the double echo broadcast tests (`crates/topos-tce-broadcast/src/tests/mod.rs`)

hadjiszs (changes requested):

crates/topos/tests/cert_delivery.rs
atanmarko: `check_certificate_delivery` checks that all nodes has delivered a certificate. `cert_delivery` checks that every subnet  should receive certificates that has cross chain transaction targeting it

Freyskeyd: This method is covering the delivery on only one target subnet and test the full delivery.
Maybe we can refactor some parts to avoid code duplication, but they don't test the same thing.
Also, we could rename this method to `assert_certificate_full_delivery`?

crates/topos/tests/cert_delivery.rs
```diff
@@ -309,3 +316,179 @@ async fn cert_delivery() {
...
+#[case(5usize)]
+#[case(9usize)]
+#[test_log::test(tokio::test)]
+#[trace]
+#[timeout(Duration::from_secs(20))]
+#[serial]
+async fn push_and_deliver_cert(
+    #[case] number_of_nodes: usize,
+) -> Result<(), Box<dyn std::error::Error>> {
+    run_assert_certificate_full_delivery(number_of_nodes, 5).await
```
hadjiszs, line 493:
```suggestion
    run_assert_certificate_full_delivery(number_of_nodes, Duration::from_secs(5)).await
```

crates/topos/Cargo.toml
```diff
@@ -69,6 +69,7 @@ rstest = { workspace = true, features = ["async-timeout"] }
 tempfile = "3.8.0"
 predicates = "3.0.3"
 sysinfo = "0.29.11"
+serial_test = {version = "0.9.0"}
```
hadjiszs, line 72: I see it introduced just yesterday, any drawback?

atanmarko: No drawbacks, prevents too many sockets opened when test are run in parallel

hadjiszs (approved):

PR #401, opened 2023-12-11
feat: add no-edge-process flag to node init

gruberb:
# Description

There are use cases for which we don't rely on `polygon-edge`. If we don't need it, we can run `node up --no-edge-process`. However, when creating a new `config` setup, we always assume `polygon-edge` is present. In the case of a `solo-sequencer`, we want to supply a fixed `validator.key` for example, and don't need or want `polygon-edge` in the process. 

Fixes [# (issue)](https://linear.app/toposware/issue/TP-817/give-the-option-to-run-node-init-without-polygon-edge)

### New feature 

Adding a `--no-edge-process` flag to `node init`.

## PR Checklist:

### DIFF
--- a/crates/topos/src/components/node/commands/init.rs
+++ b/crates/topos/src/components/node/commands/init.rs
@@ -22,4 +22,9 @@ pub struct Init {
     /// If omitted, the local FS secrets manager is used
     #[arg(long, env = "TOPOS_SECRETS_MANAGER")]
     pub secrets_config: Option<String>,
+
+    /// For certain use cases, we manually provide private keys to a running node, and don't want to
+    /// rely on polygon-edge during runtime. Example: A sequencer which runs for an external EVM chain
+    #[arg(long, env = "TOPOS_NO_EDGE_PROCESS", action)]
+    pub no_edge_process: bool,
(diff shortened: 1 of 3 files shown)


dvdplm (approved):
LGTM

I do think that the API would have been cleaner if we had flipped the logic, to have `--with-edge` rather than `--no-something-something`. That choice was taken way back though and is a bit messy to revert so not going to complain about it now.

Another tak-shave-y consideration is naming: perhaps it would have been wise to avoid naming "Edge" and instead pick something generic, like `chain-client` or `subnet-node` or something that reflects the role it plays in topos, rather than naming a specific implementation of evm chain.

gruberb:
> Another tak-shave-y consideration is naming: perhaps it would have been wise to avoid naming "Edge" and instead pick something generic, like chain-client or subnet-node or something that reflects the role it plays in topos, rather than naming a specific implementation of evm chain.

Agree. But part of me also thinks that we can very explicit of what this command does (we just have `edge` for now, and probably for a longer time?). So once this is changing, we can adapt the naming again. I rather be precise, and change it for when we want to expand.

PR #404, opened 2023-12-13
chore: adding storage context on startup

Freyskeyd:
# Description

This PR aims to add some extra context when node starts, mostly displaying the number of certificates synced, the pending/precedence pool certificates number.

```
INFO topos_tce: Storage initialized with 0 certificates delivered, 0 pending certificates and 0 certificates in the precedence pool
```
## PR Checklist:

### DIFF
--- a/crates/topos-tce-storage/src/fullnode/mod.rs
+++ b/crates/topos-tce-storage/src/fullnode/mod.rs
@@ -227,6 +227,10 @@ impl WriteStore for FullNodeStore {
 }
 
 impl ReadStore for FullNodeStore {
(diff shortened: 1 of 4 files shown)


dvdplm (approved):
LGTM

A word of warning though: counting rows this way is not cheap for big rocksdb tables and this will need to be rewritten at some point. For now it's fine and a good addition.

Freyskeyd:
@dvdplm Yes, it is mainly added to debug the devnet. If we want to add this for long term, we may rely on some properties or another storage.

PR #405, opened 2023-12-14
feat: adding filter on message when non validator

Freyskeyd:
# Description

In order to prevent many non validator's messages to be published on the gossip topics, I introduced an update where the message is filtered when the `DoubleEcho` is producing events.

For `validator` nothing changed, for `FullNode` and `Sentry` node, their `Echo` and `Ready` messages are filtered, they still produce `Gossip` messages tho.

It doesn't mean that a `FullNode` will not propagate messages from validators, it only prevents a non validator to publish messages that will be ignored by others. `FullNode` still consume `Echo` and `Ready` coming from validators and build their state same as before.

## PR Checklist:


crates/topos-tce/src/app_context/protocol.rs
```diff
@@ -83,10 +83,16 @@ impl AppContext {
...
-            evt => {
-                debug!("Unhandled event: {:?}", evt);
+            ProtocolEvents::BroadcastFailed { certificate_id } => {
+                warn!("Broadcast failed for certificate {certificate_id}")
+            }
+            ProtocolEvents::AlreadyDelivered { certificate_id } => {
+                info!("Certificate {certificate_id} already delivered")
+            }
+            ProtocolEvents::Die => {
+                error!("The DoubleEcho unexpectedly died, this is unrecoverable")
```
dvdplm, line 93: Should the process halt here? So `panic!` instead? Or…?

dvdplm (approved):
LGTM

I would copy the (excellent) PR description into the module docs for `DoubleEcho`.

crates/topos-tce/src/app_context/protocol.rs
Freyskeyd: Not really, The `Die` event is misleading I think, it's a leftover that I can need to be removed or refactor, but I hadn't time to dig deeper.

PR #415, opened 2023-12-19
feat: update tce config addresses

Freyskeyd:
# Description

This PR updates the configuration of the `TCE` by removing `libp2p_api_addr` which was originally binding and exposing the address of the node. Now the configuration uses a dedicated `P2PConfig` to hold specific p2p config, the two first parameters are:

- `listen_addresses` which accept a list of `multiaddr` that the node will listen on for new incoming connection
- `public_addresses` which define the list of addresses that are publicly exposed by the node and will be advertised by the `identify` protocol.

## Future breaking change

- The `libp2p_api_addr` is deprecated

## Example

example of a new configuration:

```toml
[base]
name = "local"
role = "validator"
subnet = "topos"

[edge]

[tce]
db-path = "/tmp/data/tce_rocksdb"
graphql-api-addr = "0.0.0.0:4030"
grpc-api-addr = "0.0.0.0:1340"
metrics-api-addr = "0.0.0.0:3000"

[tce.p2p]
public-addresses = ["/ip4/0.0.0.0/tcp/9090", "/dns/mysupernode.org/tcp/1234"]
listen-addresses = ["/ip4/0.0.0.0/tcp/9090"]
```

## PR Checklist:


crates/topos-p2p/src/network.rs
```diff
@@ -79,14 +79,14 @@ impl<'a> NetworkBuilder<'a> {
...
-        self.exposed_addresses = Some(addr);
+    pub fn advertised_addresses(mut self, addresses: Vec<Multiaddr>) -> Self {
+        self.advertised_addresses = Some(addresses);
 
         self
     }
 
-    pub fn listen_addr(mut self, addr: Multiaddr) -> Self {
-        self.listen_addr = Some(addr);
+    pub fn listen_addresses(mut self, addresses: Vec<Multiaddr>) -> Self {
```
atanmarko, line 88: Maybe `listening_addresses`?

crates/topos-p2p/src/network.rs
```diff
@@ -171,6 +171,21 @@ impl<'a> NetworkBuilder<'a> {
 
         let grpc_over_p2p = GrpcOverP2P::new(command_sender.clone());
 
+        let listen_addr = self
```
atanmarko, line 174:
```suggestion
        let listening_addrs = self
```

crates/topos-p2p/src/network.rs
```diff
@@ -171,6 +171,21 @@ impl<'a> NetworkBuilder<'a> {
...
+            .listen_addresses
+            .take()
+            .expect("Node cannot listen for connection on empty address");
```
atanmarko, line 177:
```suggestion
            .expect("Node requires one or more listening addresses provided");
```

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -62,18 +61,19 @@ impl Runtime {
...
-        let addr = self.listening_on.clone();
-        if let Err(error) = self.swarm.listen_on(addr) {
-            error!(
-                "Couldn't start listening on {} because of {error:?}",
-                self.listening_on
-            );
+        let addresses = self.listening_on.clone();
+        for addr in addresses {
+            if let Err(error) = self.swarm.listen_on(addr.clone()) {
+                error!("Couldn't start listening on {} because of {error:?}", addr);
```
atanmarko, line 73:
```suggestion
        for addr in self.listening_on {
            if let Err(error) = self.swarm.listen_on(addr.clone()) {
                error!("Couldn't start listening on {} because of {error:?}", addr);
```

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -116,7 +116,13 @@ impl Runtime {
                             let key = Key::new(&self.local_peer_id.to_string());
                             addr_query_id = if let Ok(query_id_record) =
                                 self.swarm.behaviour_mut().discovery.inner.put_record(
-                                    Record::new(key, self.addresses.to_vec()),
+                                    Record::new(
+                                        key,
+                                        self.advertised_addresses
+                                            .first()
+                                            .map(Multiaddr::to_vec)
+                                            .expect("No Addresses to advertize"),
```
atanmarko, line 124: We should return error here not crash

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -172,7 +178,13 @@ impl Runtime {
                                 let key = Key::new(&self.local_peer_id.to_string());
                                 if let Ok(query_id_record) =
                                     self.swarm.behaviour_mut().discovery.inner.put_record(
-                                        Record::new(key, self.addresses.to_vec()),
+                                        Record::new(
+                                            key,
+                                            self.advertised_addresses
+                                                .first()
+                                                .map(Multiaddr::to_vec)
+                                                .expect("No Addresses to advertize"),
```
atanmarko, line 186: Here also, return error

crates/topos-p2p/src/tests/dht.rs
```diff
@@ -40,7 +41,14 @@ async fn put_value_in_dht() {
...
         .inner
         .put_record(
-            Record::new(input_key.clone(), runtime.addresses.to_vec()),
+            Record::new(
+                input_key.clone(),
+                runtime
+                    .advertised_addresses
+                    .first()
+                    .map(Multiaddr::to_vec)
+                    .unwrap(),
```
atanmarko, line 50: No unwrap please

crates/topos-p2p/src/network.rs
```diff
@@ -39,8 +39,8 @@ const TWO_HOURS: Duration = Duration::from_secs(60 * 60 * 2);
 pub struct NetworkBuilder<'a> {
     discovery_protocol: Option<&'static str>,
     peer_key: Option<Keypair>,
-    listen_addr: Option<Multiaddr>,
-    exposed_addresses: Option<Multiaddr>,
+    listen_addresses: Option<Vec<Multiaddr>>,
+    advertised_addresses: Option<Vec<Multiaddr>>,
```
atanmarko, line 43: To be consistent I think we should use `listening_addresses` and `advertised_addresses` or `listen_addresses` and `advertise_addresses`. First one sounds better to me

crates/topos-p2p/src/tests/dht.rs
Freyskeyd: It is in tests and the `advertised_addresses` are defined just above so it can't fail.

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -62,16 +54,24 @@ impl Runtime {
 
         self.bootstrapped = true;
 
-        self.swarm.add_external_address(self.addresses.clone());
+        debug!("Added external addresses: {:?}", self.public_addresses);
```
hadjiszs, line 57:
```suggestion
        debug!("Added public addresses: {:?}", self.public_addresses);
```

not sure as I see that _public_ addresses are _external_ addresses for the swarm

crates/topos/src/components/node/services/process.rs
```diff
@@ -78,14 +79,29 @@ pub(crate) fn spawn_sequencer_process(
...
     genesis: Genesis,
     shutdown: (CancellationToken, mpsc::Sender<()>),
 ) -> JoinHandle<Result<(), Errors>> {
     let validators = genesis.validators().expect("Cannot parse validators");
     let tce_params = ReliableBroadcastParams::new(validators.len());
 
+    if let Some(socket) = config.libp2p_api_addr {
+        warn!(
+            "`libp2p_api_addr` is deprecated in favor of `listen_addresses` and \
+             `advertised_addresses` and will be removed in the next version. In order to keep \
```
hadjiszs, line 93:
```suggestion
             `public_addresses` and will be removed in the next version. In order to keep \
```

crates/topos-p2p/src/network.rs
```diff
@@ -171,6 +171,21 @@ impl<'a> NetworkBuilder<'a> {
...
+
+        let public_addresses = if let Some(addresses) = self.public_addresses.take() {
+            if addresses.is_empty() {
+                listen_addr.clone()
+            } else {
+                addresses
+            }
+        } else {
+            listen_addr.clone()
+        };
```
hadjiszs, line 187: maybe something in the likes:

```suggestion
        let public_addresses = match self.public_addresses.take() {
            None | Some(addresses) if addresses.is_empty() => listen_addr.clone()
            Some(addresses) => addresses
        };
```

might be easier to follow what is what

Freyskeyd: This will not compile as `addresses` is not present in the `None`...

hadjiszs: My bad, but you got the idea I guess, just need to separate the `None` case then no? Nitpick but it feels clearer

atanmarko (approved):

PR #416, opened 2023-12-19
chore: updating double echo for devnet

Freyskeyd:
# Description

Updating double echo for devnet debug

## PR Checklist:

### DIFF
--- a/crates/topos-tce-broadcast/src/double_echo/mod.rs
+++ b/crates/topos-tce-broadcast/src/double_echo/mod.rs
@@ -24,6 +24,7 @@ use topos_core::{
     uci::{Certificate, CertificateId},
 };
 use topos_crypto::messages::{MessageSigner, Signature};
+use topos_tce_storage::store::ReadStore;
 use topos_tce_storage::types::CertificateDeliveredWithPositions;
 use topos_tce_storage::validator::ValidatorStore;
 use tracing::{debug, error, info, warn};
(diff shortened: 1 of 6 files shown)


crates/topos-tce-broadcast/src/double_echo/mod.rs
```diff
@@ -227,23 +223,29 @@ impl DoubleEcho {
...
-        }
-
-        if self
-            .delivery_state_for_new_cert(cert, origin)
-            .await
-            .is_none()
-        {
-            error!("Ill-formed samples");
-            _ = self.event_sender.try_send(ProtocolEvents::Die);
+                .unwrap(),
```
dvdplm, line 232: I know this didn't change in this PR, but why even use `try_send` is we `unwrap` it right afterwards?

crates/topos-tce-broadcast/src/double_echo/mod.rs
```diff
@@ -190,10 +188,8 @@ impl DoubleEcho {
                     }
                 }
 
-                Some((certificate_id, status)) = task_completion.recv() => {
-                    if let TaskStatus::Success = status {
-                        self.delivered_certificates.insert(certificate_id);
-                    }
+                Some((_certificate_id, _status)) = task_completion.recv() => {
+
                 }
```
dvdplm, line 193: Why keep this at all? If it's a temporary change, please add a comment.

dvdplm (approved):
I think this is some sort of temporary change to make debugging easier? If so it would be good to state that in the PR description.

atanmarko (approved):

PR #417, opened 2023-12-19
chore: update signal handle by tce

Freyskeyd:
# Description

Update the `TCE` to catch `SIGTERM` signal coming from the OS

## PR Checklist:


dvdplm (approved):
We should have a test for this (this == shutdown), but it'd have to be an e2e test I believe.

Freyskeyd:
@dvdplm Yes I don't know if it is really possible to unit tests the signal handling on that part, sending the SIGTERM to the current PID would just kill the test process I think?

PR #418, opened 2023-12-20
feat: update double echo to use pending CF

Freyskeyd:
# Description

This PR is the final one regarding the stabilization of the `soft-fork` playbook for `devnet`.

It contains various fixes and changes that allow the network to be upgraded without the need of purging it, allowing the continuity of the delivery while upgrading. This PR biggest change is that the `broadcast` is actively asking for certificates, instead of receiving them from a channel. It allows us to stack incoming certificates in the Storage and process them at node pace. When shutting down, the broadcast will restart from the next certificate still in pending in the Storage.

### `topos`

- Update on the `node up` command, the parsing of the `boot_peers` from genesis is needed in order to define the `is-bootnode` config value as soon as possible.
- The `is-bootnode` is a config value but is not available in the config file. Mostly to prevent anyone of defining `is-bootnode = true` directly, relying solely on the `genesis` file.

### `topos-p2p`

- Uses a `is-bootnode` from configuration in place of the old value
- Add new logs in order to detect corner cases such as `kademelia` error when bootstrapping

### `topos-tce-broadcast`

- Refactor how the `task_manager` is handling shutdown
- The `task_manager` will actively request for new `pending` certificates on interval (currently 1sec), improvements need to be made on this, it'll be part of another PR.
- Add logs to better understand the flow of a certificate or message inside the task manager

### `topos-tce-storage`

- Add a new variant on the `InternalStorageError` to catch when a certificate is `CertificateAlreadyPending` 
	- The `PendingResult` also expose a `AlreadyPending`
- Add a new method on `Map` trait to `iter_at` a specific index, implemented in `DBColumn`
- `ValidatorStore` exposes a new `get_next_pending_certificates` allowing to get the next `N` certificates in pending
- Logs have been added to better understand the flow of the certificate inside the storage
- Inserting a pending certificate that is already present will return the newly define `CertificateAlreadyPending`
- The `ValidatorStore` implementation of `insert_certificate_delivered` will now correctly remove the child certificate from `precedence_pool` after adding it to `pending_pool`
- `ValidatorPendingTables` is properly fetching the current `pending_id` from the storage when starting, if not, falling back as expected on `1`

### `topos-tce`

- Dump values of the storage when shutting down to better understand the final state of the node before it stops
- Improve various logs in different places (`app_context/api` and `app_context/network` mostly)

## PR Checklist:

### DIFF
--- a/crates/topos-tce-broadcast/src/task_manager/mod.rs
+++ b/crates/topos-tce-broadcast/src/task_manager/mod.rs
@@ -103,7 +103,7 @@ impl TaskManager {
                             for (pending_id, certificate) in pendings {
                                 debug!("Creating task for pending certificate {} if needed", certificate.id);
                                 self.create_task(&certificate, true);
-                                self.latest_pending_id = pending_id + 1;
+                                self.latest_pending_id = pending_id;
                             }
                         }
                         Err(error) => {
--- a/crates/topos-tce-broadcast/src/task_manager/mod.rs
+++ b/crates/topos-tce-broadcast/src/task_manager/mod.rs
@@ -27,6 +27,7 @@ use tracing::warn;
 
 pub mod task;
 
+use crate::constant::PENDING_LIMIT_PER_REQUEST_TO_STORAGE;
 use crate::double_echo::broadcast_state::BroadcastState;
 use crate::sampler::SubscriptionsView;
 use crate::DoubleEchoCommand;
@@ -97,7 +98,7 @@ impl TaskManager {
 
                 _ = interval.tick() => {
                     debug!("Checking for next pending_certificates");
-                    match self.validator_store.get_next_pending_certificates(&self.latest_pending_id, 1000) {
+                    match self.validator_store.get_next_pending_certificates(&self.latest_pending_id, *PENDING_LIMIT_PER_REQUEST_TO_STORAGE) {
                         Ok(pendings) => {
                             debug!("Received {} pending certificates", pendings.len());
                             for (pending_id, certificate) in pendings {
@@ -146,11 +147,11 @@ impl TaskManager {
                 _ = shutdown_receiver.cancelled() => {
                     warn!("Task Manager shutting down");
 
-                    warn!("There is still {} active tasks", self.tasks.len());
+                    warn!("There are still {} active tasks", self.tasks.len());
                     if !self.tasks.is_empty() {
-                        debug!("Certificate still in broadcast: {:?}", self.tasks.keys());
+                        debug!("Certificates still in broadcast: {:?}", self.tasks.keys());
                     }
-                    warn!("There is still {} buffered messages", self.buffered_messages.len());
+                    warn!("There are still {} buffered messages", self.buffered_messages.len());
                     for task in self.tasks.iter() {
                         task.1.shutdown_sender.send(()).await.unwrap();
                     }
--- a/crates/topos-tce-broadcast/src/task_manager/mod.rs
+++ b/crates/topos-tce-broadcast/src/task_manager/mod.rs
@@ -56,7 +56,6 @@ pub struct TaskManager {
     pub broadcast_sender: broadcast::Sender<CertificateDeliveredWithPositions>,
 
     pub latest_pending_id: PendingCertificateId,
-    pub precedence: HashMap<CertificateId, Task>,
 }
 
 impl TaskManager {
@@ -86,7 +85,6 @@ impl TaskManager {
             validator_store,
             broadcast_sender,
             latest_pending_id: 0,
-            precedence: HashMap::new(),
         }
     }
 
@@ -98,9 +96,12 @@ impl TaskManager {
                 biased;
 
                 _ = interval.tick() => {
+                    debug!("Checking for next pending_certificates");
                     match self.validator_store.get_next_pending_certificates(&self.latest_pending_id, 1000) {
                         Ok(pendings) => {
+                            debug!("Received {} pending certificates", pendings.len());
                             for (pending_id, certificate) in pendings {
+                                debug!("Creating task for pending certificate {} if needed", certificate.id);
                                 self.create_task(&certificate, true);
                                 self.latest_pending_id = pending_id + 1;
                             }
@@ -133,24 +134,12 @@ impl TaskManager {
 
                 Some((certificate_id, status)) = self.running_tasks.next() => {
                     if let TaskStatus::Success = status {
+                        debug!("Task for certificate {} finished successfully", certificate_id);
                         self.tasks.remove(&certificate_id);
                         DOUBLE_ECHO_ACTIVE_TASKS_COUNT.dec();
                         let _ = self.task_completion_sender.send((certificate_id, status)).await;
-                        if let Some(task) = self.precedence.remove(&certificate_id) {
-                            if let Some(context) = self.tasks.get(&task.certificate_id) {
-
-                                let certificate_id= task.certificate_id;
-                                Self::start_task(
-                                    &self.running_tasks,
-                                    task,
-                                    context.sink.clone(),
-                                    self.buffered_messages.remove(&certificate_id),
-                                    false
-                                );
-                            }
-
-
-                        }
+                    } else {
+                        debug!("Task for certificate {} finished unsuccessfully", certificate_id);
                     }
                 }
 
@@ -239,7 +228,6 @@ impl TaskManager {
                          certificate {} is not available yet",
                         cert.id, cert.prev_id
                     );
-                    self.precedence.insert(cert.prev_id, task);
                 }
                 entry.insert(task_context);
             }
--- a/crates/topos-tce-broadcast/src/task_manager/mod.rs
+++ b/crates/topos-tce-broadcast/src/task_manager/mod.rs
@@ -95,7 +95,7 @@ impl TaskManager {
             tokio::select! {
                 biased;
 
-                _ = interval.tick() => {
+                _ = interval.tick(), if !shutdown_receiver.is_cancelled() => {
                     debug!("Checking for next pending_certificates");
                     match self.validator_store.get_next_pending_certificates(&self.latest_pending_id, 1000) {
                         Ok(pendings) => {
@@ -103,7 +103,7 @@ impl TaskManager {
                             for (pending_id, certificate) in pendings {
                                 debug!("Creating task for pending certificate {} if needed", certificate.id);
                                 self.create_task(&certificate, true);
-                                self.latest_pending_id = pending_id;
+                                self.latest_pending_id = pending_id + 1;
                             }
                         }
                         Err(error) => {
@@ -123,10 +123,13 @@ impl TaskManager {
                                     .push(msg);
                             };
                         }
-                        DoubleEchoCommand::Broadcast { ref cert, need_gossip } => {
+                        DoubleEchoCommand::Broadcast { ref cert, need_gossip } if !shutdown_receiver.is_cancelled() => {
                             debug!("Received broadcast message for certificate {} ", cert.id);
 
                             self.create_task(cert, need_gossip)
+                        }
+                        DoubleEchoCommand::Broadcast{..} => {
+                            // Ignored due to shutdown
                         }
                     }
                 }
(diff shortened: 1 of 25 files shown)


crates/topos-tce-broadcast/src/task_manager/mod.rs
```diff
@@ -109,77 +125,32 @@ impl TaskManager {
...
+                    } else {
+                        debug!("Task for certificate {} finished unsuccessfully", certificate_id);
                     }
                 }
 
-                _ = shutdown_receiver.recv() => {
+                _ = shutdown_receiver.cancelled() => {
                     warn!("Task Manager shutting down");
 
+                    warn!("There is still {} active tasks", self.tasks.len());
```
dvdplm, line 149:
```suggestion
                    warn!("There are still {} active tasks", self.tasks.len());
```

crates/topos-tce-broadcast/src/task_manager/mod.rs
```diff
@@ -109,77 +125,32 @@ impl TaskManager {
...
+                    if !self.tasks.is_empty() {
+                        debug!("Certificate still in broadcast: {:?}", self.tasks.keys());
```
dvdplm, line 151:
```suggestion
                        debug!("Certificates still in broadcast: {:?}", self.tasks.keys());
```

crates/topos-tce-broadcast/src/task_manager/mod.rs
```diff
@@ -109,77 +125,32 @@ impl TaskManager {
...
+                    }
+                    warn!("There is still {} buffered messages", self.buffered_messages.len());
```
dvdplm, line 153:
```suggestion
                    warn!("There are still {} buffered messages", self.buffered_messages.len());
```

crates/topos-tce-storage/src/tests/mod.rs
```diff
@@ -354,17 +355,22 @@ async fn pending_certificate_can_be_removed(store: Arc<ValidatorStore>) {
...
+        store.insert_pending_certificate(&certificate),
+        Err(StorageError::InternalStorage(
+            crate::errors::InternalStorageError::CertificateAlreadyPending
+        ))
+    ));
+
     assert!(pending_column.get(&pending_id).is_ok());
     store.delete_pending_certificate(&pending_id).unwrap();
 
-    assert!(pending_column.iter().unwrap().next().is_some());
```
dvdplm, line 367: Was this test wrong before?

crates/topos-tce-storage/src/validator/mod.rs
```diff
@@ -118,6 +118,23 @@ impl ValidatorStore {
...
+        number: usize,
+    ) -> Result<Vec<(PendingCertificateId, Certificate)>, StorageError> {
+        debug!(
+            "Get next pending certificates from {} (max: {})",
+            from, number
+        );
+        Ok(self
+            .pending_tables
+            .pending_pool
+            .iter_at(from)?
```
dvdplm, line 133: Not a task for this PR ofc but we should have a think about building a good benchmark for code like this so we know what performance characteristics we can expect for different DB sizes. Rocksdb iterators are quirky and sometimes surprising.

crates/topos-tce-storage/src/validator/mod.rs
```diff
@@ -213,11 +243,19 @@ impl ValidatorStore {
...
+                "Certificate {} is now in the pending pool at index: {}",
+                certificate.id, id
+            );
             Ok(Some(id))
         } else {
             self.pending_tables
                 .precedence_pool
                 .insert(&certificate.prev_id, certificate)?;
+            debug!(
+                "Certificate {} is now in the precedence pool, {} isn't delivered yet",
```
dvdplm, line 256:
```suggestion
                "Certificate {} is now in the precedence pool, because the previous cert with id {} isn't delivered yet",
```

maybe?

crates/topos-tce-storage/src/validator/mod.rs
```diff
@@ -467,12 +505,19 @@ impl WriteStore for ValidatorStore {
...
 
-        if let Ok(Some(certificate)) = self
+        if let Ok(Some(next_certificate)) = self
             .pending_tables
             .precedence_pool
             .get(&certificate.certificate.id)
         {
-            self.insert_pending_certificate(&certificate)?;
+            debug!(
+                "Delivered certificate {} unlocks {} for broadcast",
```
dvdplm, line 514:
```suggestion
                "Delivered certificate {}. This unlocks {} for broadcast",
```

crates/topos-tce-storage/src/validator/tables.rs
```diff
@@ -71,11 +76,38 @@ impl ValidatorPendingTables {
 
         let db = init_with_cfs(&path, default_options(), cfs)
             .unwrap_or_else(|_| panic!("Cannot open DB at {:?}", path));
+        let pending_pool = DBColumn::reopen(&db, cfs::PENDING_POOL);
+        let next_pending_id = {
+            let cf = pending_pool
+                .rocksdb
+                .cf_handle(cfs::PENDING_POOL)
+                .expect("Cannot get cf handle for pending pool");
```
dvdplm, line 84: I guess the idea here is that if we can't open the DB then it's better to panic and exit rather than returning an error? Do we have some sort of team-agreement on the kinds of "catastrophic errors" that should stop the node from starting? Maybe something we should add to the style guide?

crates/topos-tce-storage/src/validator/tables.rs
```diff
@@ -71,11 +76,38 @@ impl ValidatorPendingTables {
...
+            let mut pending_iterator = pending_pool.rocksdb.raw_iterator_cf(&cf);
+
+            pending_iterator.seek_to_last();
```
dvdplm, line 87: Is this expensive for a big DB? Or does rocksdb cache the last index?

crates/topos-tce-storage/src/validator/tables.rs
```diff
@@ -71,11 +76,38 @@ impl ValidatorPendingTables {
...
+                                .unwrap_or(0)
+                        })
+                        .unwrap_or(0),
+                )
+            } else {
+                AtomicU64::new(0)
+            }
+        };
+
+        next_pending_id.fetch_add(1, Ordering::Relaxed);
```
dvdplm, line 106: Are you sure about `Relaxed` is ok here? I'd have gone for `SeqCst` to be safe...

crates/topos-tce/src/app_context.rs
```diff
@@ -141,11 +143,35 @@ impl AppContext {
 
     pub async fn shutdown(&mut self) -> Result<(), Box<dyn std::error::Error>> {
         info!("Shutting down the TCE client...");
+
         self.api_client.shutdown().await?;
         self.tce_cli.shutdown().await?;
         self.gatekeeper.shutdown().await?;
         self.network_client.shutdown().await?;
```
dvdplm, line 150: Maybe we should crash here too if they fail?

dvdplm: Also: I think it would be good to have a timeout in this method, forcing shutdown after `n` seconds regardless of any subsystem being slow to shutdown. Daemons that don't shutdown properly are so annoying.

dvdplm (approved):
Lots of good micro-improvments here, good stuff.

I fixed a few typos and log messages, but the rest of my comments are questions or topics best addressed in separate PRs. Feel free to respond to question at your convenience (not a blocker to merge).

crates/topos-tce-broadcast/src/task_manager/mod.rs
```diff
@@ -66,35 +70,47 @@ impl TaskManager {
...
+    pub async fn run(mut self, mut shutdown_receiver: CancellationToken) {
+        let mut interval = tokio::time::interval(Duration::from_secs(1));
+
         loop {
             tokio::select! {
                 biased;
 
+                _ = interval.tick() => {
+                    debug!("Checking for next pending_certificates");
+                    match self.validator_store.get_next_pending_certificates(&self.latest_pending_id, 1000) {
```
gruberb, line 100: The amount (`1000`) should be a `const` somewhere, or be able to passed down. Also, where does the number come from? Doesn't the `storage` know how many are in pending, and we just pass the latest `pending_id` down?

crates/topos-tce-storage/src/validator/mod.rs
```diff
@@ -467,12 +505,19 @@ impl WriteStore for ValidatorStore {
...
+                certificate.certificate.id, next_certificate.id
+            );
+            self.insert_pending_certificate(&next_certificate)?;
```
gruberb, line 517: Why would we insert it as pending here?

gruberb (changes requested):
Just one nitpick about having a `const` for the amount of `pending_certificates` we want to fetch with each `tick`. 

Generally, I would like some explanation how the intented flow is. When we re-deploy the network, how is the `pending_certificate` help us? Do we do partial shutdowns and then the nodes pick up, and then get shutdown?

crates/topos-tce-storage/src/validator/mod.rs
Freyskeyd: A `certificate` that waits for its `parent` to be delivered is stored in the `precedence_pool` under `parent_id => certificate`. When a `certificate` is delivered, we check if we have a `child` waiting in the precedence pool, if yes, we put this child in the `pending` pool in order to be delivered.

crates/topos-tce-broadcast/src/task_manager/mod.rs
Freyskeyd: The number `1000` is a random-picked value, and having a `limit` on the method allow any other app to require only the amount of certificate in pending that they want.

crates/topos-tce-storage/src/validator/tables.rs
Freyskeyd: I think it is ok as the variable is creating in this function, what do you think?

crates/topos-tce-storage/src/tests/mod.rs
Freyskeyd: Not really, it was updated regarding the add of the `AlreadyPending` variant.

crates/topos-tce-storage/src/validator/tables.rs
Freyskeyd: We didn't agree on having such kind of logic but having the node panicking during the storage setup is ok for me, it prevents us to deal with a big error management. And basically, if we can't create the storage, the node will not properly work and should exit. We can discuss if we want a full error management, but for the sake of simplicity, I didn't wanted to introduce it in this PR.

dvdplm: I agree with you. I was just wondering if we have any sort of rule around this.

crates/topos-tce-storage/src/validator/tables.rs
dvdplm: Fair, it should be fine. We use `Relaxed` everywhere though – probably worth to sit down and review those too.

crates/topos-tce-storage/src/validator/tables.rs
Freyskeyd: I'm not sure, but if I recall correctly, `rocksdb` based its `iterator` on an implicit snapshot of the data, it uses an ordered index to target the right key, I think.

crates/topos-tce-storage/src/validator/mod.rs
Freyskeyd: I think I'll keep the PR format as I was using it really often when debug just by doing ` grep 'certA unlocks certB'`.

crates/topos-tce/src/app_context.rs
Freyskeyd: Mmmh there is two approach here, either we define that we have a timeout that kill everything internally, or we let the system management using a SIGKILL after X sec.
On a larger scope, I'm not really satisfied with the current handling of the shutdown, we should be safe to cut the service without cleanup at any time and being able to recover on startup. It may need more work on various part tho.

crates/topos-tce-storage/src/validator/mod.rs
Freyskeyd: Indeed, we should have a long term prioritization of those underlying subject that doesn't really fit in features.

gruberb (approved):
👍

crates/topos-tce/src/app_context.rs
dvdplm: > we should be safe to cut the service without cleanup at any time and being able to recover on startup

Agreed. I think it's the kind of discipline that has benefits beyond just shutdowns. It kind of forces us to write correct code.

crates/topos-tce-storage/src/validator/mod.rs
dvdplm: No strong opinion on how to capture these things, as long as they are written down somewhere in a ticket with a proper label.

PR #422, opened 2023-12-26
fix: return error from subprocesses

atanmarko:
# Description

Return topos cli exit status with respect to edge subprocess status

Fixes TP-825

## PR Checklist:

### DIFF
--- a/crates/topos/src/components/node/mod.rs
+++ b/crates/topos/src/components/node/mod.rs
@@ -270,7 +270,6 @@ pub(crate) async fn handle_command(
                     shutdown(basic_controller, shutdown_trigger, shutdown_receiver).await;
                 }
                 Some(result) = processes.next() => {
-                    info!("Terminate: {result:?}");
                     shutdown(basic_controller, shutdown_trigger, shutdown_receiver).await;
                     processes.clear();
                     match result {
@@ -286,7 +285,8 @@ pub(crate) async fn handle_command(
                             error!("Terminating with error: {e}");
                             std::process::exit(1);
                         }
-                        Err(_) => {
+                        Err(e) => {
+                            error!("Terminating with error: {e}");
                             std::process::exit(1);
                         }
                     }
--- a/crates/topos/src/components/node/mod.rs
+++ b/crates/topos/src/components/node/mod.rs
@@ -79,13 +79,30 @@ pub(crate) async fn handle_command(
                 println!("Init the node without polygon-edge process...");
             } else {
                 // Generate the Edge configuration
-                if let Ok(result) = services::process::generate_edge_config(
+                match services::process::generate_edge_config(
                     edge_path.join(BINARY_NAME),
                     node_path.clone(),
                 )
                 .await
                 {
-                    if result.is_err() {
+                    Ok(Ok(status)) => {
+                        if let Some(0) = status.code() {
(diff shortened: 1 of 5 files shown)


atanmarko:
Somehow polygon edge spawned process always returns `ExitStatus(unix_wait_status(0))` despite the fact that error happened. I have tried other commands (`ls`) and their error is correctly propagated to `topos cli` exit

crates/topos/src/components/node/mod.rs
```diff
@@ -79,13 +79,30 @@ pub(crate) async fn handle_command(
...
+                    }
+                    Ok(Err(e)) => {
+                        println!("Failed to generate edge config with error {e}");
+                        remove_dir_all(node_path).expect("failed to remove config folder");
+                        std::process::exit(1);
+                    }
+                    Err(e) => {
                         println!("Failed to generate edge config");
                         remove_dir_all(node_path).expect("failed to remove config folder");
                         std::process::exit(1);
```
Freyskeyd, line 108:
```suggestion
                    Ok(Err(e)) | Err(e) => {
                        println!("Failed to generate edge config with error {e}");
                        remove_dir_all(node_path).expect("failed to remove config folder");
                        std::process::exit(1);
```

I think you can simplify this with that

crates/topos/src/components/node/mod.rs
```diff
@@ -253,12 +270,26 @@ pub(crate) async fn handle_command(
...
+                    match result {
+                        Ok(Ok(status)) => {
+                            if let Some(0) = status.code() {
+                                info!("Terminating with success error code");
+                            } else {
+                                info!("Terminating with error status: {:?}", status);
+                                std::process::exit(1);
+                            }
+                        }
+                        Ok(Err(e)) => {
```
Freyskeyd, line 284: Same here for the shorter version

crates/topos/src/edge/mod.rs
```diff
@@ -99,8 +100,18 @@ impl CommandConfig {
 
         let (running_out, _) = tokio::join!(running, logging);
 
-        info!("The Edge process is terminated");
-        running_out
+        let exit_status = running_out?;
+
+        info!(
```
Freyskeyd, line 105: That's a lot of information, let's keep that like that and see later if it doesn't display too much things.

Freyskeyd (approved):

crates/topos/src/components/node/mod.rs
atanmarko: Tried it already, compiler gives:
```
 this expression has type `std::result::Result<std::result::Result<std::process::ExitStatus, components::node::services::process::Errors>, tokio::task::JoinError>`
...
100 |                       Ok(Err(e)) |  Err(e) => {
    |                              -          ^ expected `Errors`, found `JoinError`
```

crates/topos/src/components/node/mod.rs
atanmarko: Same compiler error as before - error must be of the same type to match such a pattern

crates/topos/src/components/node/mod.rs
Freyskeyd: Ah it is a `JoinError` ?
Maybe you can remove the `JoinHandle` and transform `generate_edge_config` into an `async` function.

PR #426, opened 2024-01-04
feat: remove the register commands macro

gruberb:
# Description

Remove not needed `RegisterCommands` macro, which was used for `GetAllSubnets` in the `Gatekeeper` client. This is not needed anymore.

Fixes TP-562

## PR Checklist:


gruberb:
Seems like we don't use `GetAllSubnets` anymore. Or am I wrong @Freyskeyd?

Freyskeyd (approved):
The `gatekeeper` is useless for now as we do not handle the `validators` in it. We'll need to re-add some part when working on it later.

PR #433, opened 2024-01-19
fix: remove an unused channel that was locking the broadcast

Freyskeyd:
# Description

The broadcast on `devnet-1` was stuck after some time.
It was because an unused channel was filling up by un-consumed messages and the N+1 message was waiting for slot that was never released.

## PR Checklist:


gruberb (approved):
👍

PR #435, opened 2024-01-22
refactor: graphql types to differentiate inputs

Freyskeyd:
# Description

This PR prepares the work to add the position and expose more fields to GraphQL.
It updates the different `input` such as `SourceStreamPosition` to `SourceStreamPositionInput`.

## Breaking changes

BREAKING CHANGE: This break the `0.0.10` GraphQL API definition due to the rename of some fields.

## PR Checklist:


crates/topos-core/src/uci/subnet_id.rs
```diff
@@ -88,6 +88,16 @@ impl FromStr for SubnetId {
     }
 }
 
+impl PartialEq<[u8]> for SubnetId {
+    fn eq(&self, other: &[u8]) -> bool {
+        if let Ok(current) = Self::try_from(other) {
+            self.as_array().eq(current.as_array())
```
dvdplm, line 94: What's the advantage of doing this? As opposed to `self.as_array() == other`? Why convert `other` to a SubnetId first?

dvdplm (approved):

gruberb (approved):

crates/topos-core/src/uci/subnet_id.rs
Freyskeyd: It validates the format of the subnet id provided as `[u8]` which can contains extra `0x` that need to be removed.

PR #438, opened 2024-01-23
chore: remove mention of topos-api from coverage YAML

dvdplm:
Cleanup after #436.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -45,7 +45,6 @@ jobs:
             -t lcov \
             --branch \
             --ignore-not-existing \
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yml
```diff
@@ -45,7 +45,6 @@ jobs:
             -t lcov \
             --branch \
             --ignore-not-existing \
-            --ignore 'crates/topos-api/src/grpc/generated/*' \
```
Freyskeyd, line 48: Ah good catch but we may need to keep this but change the line to `crates/topos-core/src/api/grpc/generated/*`

Freyskeyd (approved):

PR #443, opened 2024-01-29
feat: introduce topos-config crate

Freyskeyd:
# Description

This PR introduces a new crate `topos-config` which will provide struct for the different configuration across the various crates.

## PR Checklist:


crates/topos-config/src/lib.rs
```diff
@@ -77,7 +70,7 @@ pub(crate) fn load_config<T: Config>(node_path: &Path, command: Option<NodeComma
     }
 }
 
-pub(crate) fn insert_into_toml<T: Config>(config_toml: &mut toml::Table, config: T) {
+pub fn insert_into_toml<T: Config>(config_toml: &mut toml::Table, config: T) {
```
dvdplm, line 73: Why not use a `From` impl to do this?

crates/topos-config/src/node.rs
```diff
@@ -21,29 +20,32 @@ pub enum NodeRole {
...
+    pub base: BaseConfig,
+    pub tce: Option<TceConfig>,
+    pub sequencer: Option<SequencerConfig>,
+    pub edge: Option<EdgeConfig>,
 }
 
 impl NodeConfig {
-    pub fn new(home: &Path, cmd: Option<NodeCommands>) -> Self {
-        let base = load_config::<BaseConfig>(home, cmd);
+    pub fn new<T: Serialize>(home: &Path, cmd: Option<T>) -> Self {
```
dvdplm, line 35: Use `S` here, for consistency? Or perhaps even better `C` or even `Cmd`?

crates/topos-config/src/tce.rs
```diff
@@ -6,22 +7,55 @@ use figment::{
...
+use crate::Config;
 use topos_p2p::{Multiaddr, PeerId};
 
 const DEFAULT_IP: std::net::Ipv4Addr = std::net::Ipv4Addr::new(0, 0, 0, 0);
 
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Debug)]
+pub enum AuthKey {
+    Seed(Vec<u8>),
+    PrivateKey(Vec<u8>),
```
dvdplm, line 22: Aren't these known sizes, e.g. 32 bytes? Can we use an array?

crates/topos-tce/src/lib.rs
```diff
@@ -135,18 +135,18 @@ pub async fn run(
...
         .grpc_context(grpc_context)
-        .is_bootnode(config.is_bootnode)
+        .is_bootnode(config.p2p.is_bootnode)
         .build()
         .await?;
 
     debug!("Starting the p2p network");
     let network_runtime = tokio::time::timeout(
-        config.network_bootstrap_timeout,
+        Duration::from_secs(config.network_bootstrap_timeout),
```
dvdplm, line 149: Why is it better to timeout here as opposed to retry periodically until it can connect?

crates/topos/src/components/node/services/process.rs
```diff
@@ -113,29 +111,12 @@ pub(crate) fn spawn_tce_process(
...
-        graphql_api_addr: config.graphql_api_addr,
-        metrics_api_addr: config.metrics_api_addr,
-        storage: StorageConfiguration::RocksDB(Some(config.db_path)),
-        network_bootstrap_timeout: Duration::from_secs(config.network_bootstrap_timeout),
-        minimum_cluster_size: config
-            .minimum_tce_cluster_size
-            .unwrap_or(NetworkConfig::MINIMUM_CLUSTER_SIZE),
-        version: env!("TOPOS_VERSION"),
-    };
+    config.version = env!("TOPOS_VERSION");
```
dvdplm, line 114: How is this ENV var set? What value does `config.version` take if it's not set? We use figment elsewhere to define a hierarchy of config value sources. Can we do the same for the version? Ideally I'd want this to use the crate version of the `topos` crate + the short hash of the build, and not have a user-inputted value at all.

crates/topos-config/src/node.rs
```diff
@@ -21,29 +20,32 @@ pub enum NodeRole {
...
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone)]
-pub(crate) struct NodeConfig {
-    pub(crate) base: BaseConfig,
-    pub(crate) tce: Option<TceConfig>,
-    pub(crate) sequencer: Option<SequencerConfig>,
-    pub(crate) edge: Option<EdgeConfig>,
+#[derive(Serialize)]
+struct Noop {}
```
dvdplm, line 24: It's a bit annoying to have to add this type. Wouldn't `()` work? If we indeed have to have this, a better name would be `NoopCommand` or something that explains what it is.

crates/topos-config/src/lib.rs
Freyskeyd: Its not new code, so I didn't touch the design.
But feel free to suggest code if you want.

crates/topos-config/src/tce.rs
Freyskeyd: As above, it is not new code. And no it's not guaranteed that they're 32 bytes if I recall correctly, but I'm not the author of this part, maybe @hadjiszs has more information

crates/topos/src/components/node/services/process.rs
Freyskeyd: the version is computed in the `build.rs`

crates/topos-tce/src/lib.rs
Freyskeyd: It'll maybe change soon with the work on the benchmarks, but the goal of this PR is not to refactor how the network behave

crates/topos-config/src/lib.rs
dvdplm: Thing is: you're making this method public, which means we have to support it (or bear the burden of breaking APIs), so that marks a transition when things like bad names start to matter a bit more.

dvdplm: If using `From` is annoying here, consider removing the method and just copy the flattening stuff into where it's used.

crates/topos-config/src/tce.rs
dvdplm: 🤷 Is it important if it's new code or not? I think not.

If we have doubts about the sizes of the seed and private key, perhaps that's an indication that the types used are not expressive enough.

crates/topos-config/src/lib.rs
Freyskeyd: @dvdplm I don't see how you could implement `From` in this situation. I will see if I can remove or refactor this.

crates/topos-config/src/tce.rs
hadjiszs: Seed can be any size, then `AuthKey::PrivateKey` was with abstracted size to handle different private keys (ed25519 ones of 34bytes, or secp256k1 of 32bytes) as we were using different curve between `PeerId` and `ValidatorId` before

Right now I re-checked and this is used only with secp256k1 private key so always 32 bytes for now

I can create one ticket to refine these types with more explicit types, we could work on it while switching to the edge alternative as these structs are initially fed from the key files generated by edge

dvdplm (approved):

PR #447, opened 2024-02-05
chore: debug 0.0.11 synchronization

Freyskeyd:
# Description

This PR explores the issue with the synchronization of a new node. It solves the issue by refactoring a bit the calculation of diff between local and input.
The problem relied on the `prefix_iter_at` method that was introduced during the synchronization phase, but didn't work as expected in some cases. A ticket has been created to fix or refactor it, in the meantime, the synchronization process is falling back to the `skip` method.

## PR Checklist:

### DIFF
--- a/crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
+++ b/crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
@@ -230,13 +230,20 @@ impl BroadcastState {
 
     fn reached_delivery_threshold(&self) -> bool {
         // If reached the delivery threshold, I can deliver
-        match self
+        let delivery_threshold = match self
             .subscriptions_view
             .network_size
             .checked_sub(self.subscriptions_view.ready.len())
         {
             Some(consumed) => consumed >= self.delivery_threshold,
             None => false,
-        }
+        };
+
+        debug!(
+            "📝 Certificate {} reached Delivery threshold: {}",
+            &self.certificate.id, delivery_threshold
+        );
+
+        delivery_threshold
     }
 }
--- a/crates/topos-tce-storage/src/tests/checkpoints.rs
+++ b/crates/topos-tce-storage/src/tests/checkpoints.rs
@@ -54,7 +54,7 @@ async fn get_checkpoint_diff_with_no_input(store: Arc<ValidatorStore>) {
     }
 
     let checkpoint = store
-        .get_checkpoint_diff(vec![])
+        .get_checkpoint_diff(&[])
         .unwrap()
         .into_iter()
         .map(|(subnet, proofs)| {
@@ -97,7 +97,7 @@ async fn get_checkpoint_diff_with_input(store: Arc<ValidatorStore>) {
     }
 
     let checkpoint = store
-        .get_checkpoint_diff(vec![checkpoint])
+        .get_checkpoint_diff(&[checkpoint])
         .unwrap()
         .into_iter()
         .map(|(subnet, proofs)| {
--- a/crates/topos-tce-storage/src/tests/checkpoints.rs
+++ b/crates/topos-tce-storage/src/tests/checkpoints.rs
@@ -1,0 +1,123 @@
+use std::{collections::HashMap, sync::Arc};
+
+use rstest::rstest;
+use topos_core::uci::SubnetId;
+use topos_test_sdk::{
+    certificates::create_certificate_chain,
+    constants::{SOURCE_SUBNET_ID_1, SOURCE_SUBNET_ID_2, TARGET_SUBNET_ID_1},
+};
+
+use super::support::store;
+use crate::{
+    store::{ReadStore, WriteStore},
+    validator::ValidatorStore,
+};
+
+#[rstest]
+#[tokio::test]
+async fn get_checkpoint_for_two_subnets(store: Arc<ValidatorStore>) {
+    let certificates_a = create_certificate_chain(SOURCE_SUBNET_ID_1, &[TARGET_SUBNET_ID_1], 32);
+    let certificates_b = create_certificate_chain(SOURCE_SUBNET_ID_2, &[TARGET_SUBNET_ID_1], 24);
+
+    for cert in certificates_a {
+        _ = store.insert_certificate_delivered(&cert).await;
+    }
+
+    for cert in certificates_b {
+        _ = store.insert_certificate_delivered(&cert).await;
+    }
+
+    let checkpoint = store
+        .get_checkpoint()
+        .unwrap()
+        .into_iter()
+        .map(|(subnet, value)| (subnet, *value.position))
+        .collect::<HashMap<SubnetId, u64>>();
+
+    assert_eq!(checkpoint.len(), 2);
+    assert_eq!(*checkpoint.get(&SOURCE_SUBNET_ID_1).unwrap(), 31);
+    assert_eq!(*checkpoint.get(&SOURCE_SUBNET_ID_2).unwrap(), 23);
+}
+
+#[rstest]
+#[tokio::test]
+async fn get_checkpoint_diff_with_no_input(store: Arc<ValidatorStore>) {
+    let certificates_a = create_certificate_chain(SOURCE_SUBNET_ID_1, &[TARGET_SUBNET_ID_1], 32);
+    let certificates_b = create_certificate_chain(SOURCE_SUBNET_ID_2, &[TARGET_SUBNET_ID_1], 24);
+
+    for cert in certificates_a {
+        _ = store.insert_certificate_delivered(&cert).await;
+    }
+
+    for cert in certificates_b {
+        _ = store.insert_certificate_delivered(&cert).await;
+    }
+
+    let checkpoint = store
+        .get_checkpoint_diff(vec![])
+        .unwrap()
+        .into_iter()
+        .map(|(subnet, proofs)| {
+            (
+                subnet,
+                proofs
+                    .iter()
+                    .map(|proof| *proof.delivery_position.position)
+                    .collect::<Vec<_>>(),
+            )
+        })
+        .collect::<HashMap<SubnetId, _>>();
+
+    assert_eq!(checkpoint.len(), 2);
+    assert_eq!(
+        *checkpoint.get(&SOURCE_SUBNET_ID_1).unwrap(),
+        (0..=31).collect::<Vec<_>>()
+    );
+    assert_eq!(
+        *checkpoint.get(&SOURCE_SUBNET_ID_2).unwrap(),
+        (0..=23).collect::<Vec<_>>()
+    );
+}
+
+#[rstest]
+#[tokio::test]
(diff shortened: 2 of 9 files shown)


crates/topos-tce-storage/src/tests/checkpoints.rs
```diff
@@ -0,0 +1,123 @@
...
+    for cert in certificates_a {
+        _ = store.insert_certificate_delivered(&cert).await;
+    }
+
+    for cert in certificates_b {
+        _ = store.insert_certificate_delivered(&cert).await;
+    }
+
+    let checkpoint = store
+        .get_checkpoint_diff(vec![checkpoint])
```
atanmarko, line 100: Maybe `get_checkpoint_diff` could take references as a function - no need then to clone whole vectors?

atanmarko (approved):

crates/topos-tce-synchronizer/src/checkpoints_collector/mod.rs
```diff
@@ -184,11 +189,11 @@ impl CheckpointSynchronizer {
         diff: HashMap<SubnetId, Vec<ProofOfDelivery>>,
     ) -> Result<Vec<Vec<CertificateId>>, SyncError> {
         let mut certs: HashSet<CertificateId> = HashSet::new();
-        for (_subnet, proofs) in diff {
+        for (subnet, proofs) in diff {
             let len = proofs.len();
             let unverified_certs = self.store.insert_unverified_proofs(proofs)?;
 
-            debug!("Persist {} unverified proofs", len);
+            debug!("Persist {} unverified proofs for {}", len, subnet);
```
hadjiszs, line 196:
```suggestion
            debug!("Persist {} unverified proof of delivery for {}", len, subnet);
```

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
```diff
@@ -230,13 +230,20 @@ impl BroadcastState {
...
         {
             Some(consumed) => consumed >= self.delivery_threshold,
             None => false,
-        }
+        };
+
+        debug!(
+            "📝 Certificate {} reached Delivery threshold: {}",
+            &self.certificate.id, delivery_threshold
+        );
```
hadjiszs, line 245: This one will print out often and redundant with logs in `update_status` no? Maybe to remove or to put it as `trace`?

hadjiszs (approved):

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
Freyskeyd: It matches what we have for `echo` and `ready` threshold, when the certificate is delivered, even if we have more message coming, they're ignored

crates/topos-tce-storage/src/tests/checkpoints.rs
Freyskeyd: I updated the `get_checkpoint_diff` to take a ref

crates/topos-tce-storage/src/tests/checkpoints.rs
```diff
@@ -0,0 +1,123 @@
...
+    );
+    assert_eq!(
+        *checkpoint.get(&SOURCE_SUBNET_ID_2).unwrap(),
+        (0..=23).collect::<Vec<_>>()
+    );
+}
+
+#[rstest]
+#[tokio::test]
+async fn get_checkpoint_diff_with_input(store: Arc<ValidatorStore>) {
```
dvdplm, line 84: I do not understand what this test actually does, can you explain please? The part I don't get is what the "input" part is. The only difference between this test and the "no_input" version seems to be that this one does `.get(20)` and clones the proof of delivery, but I don't see how that maps to "with input".

crates/topos-tce-storage/src/tests/checkpoints.rs
```diff
@@ -0,0 +1,123 @@
...
+    assert_eq!(checkpoint.len(), 2);
+    assert_eq!(
+        *checkpoint.get(&SOURCE_SUBNET_ID_1).unwrap(),
+        (21..=31).collect::<Vec<_>>()
+    );
+    assert_eq!(
+        *checkpoint.get(&SOURCE_SUBNET_ID_2).unwrap(),
+        (0..=23).collect::<Vec<_>>()
+    );
+}
```
dvdplm, line 123: It would be good to have some more tests to make sure the code behaves the way we expect when things have gone wrong. Like, missing certificates, conflicting chains (some TCE node considers something delivered and some don't?) and other such unexpected scenarios. Can be different PR(s) too.

crates/topos-tce-storage/src/validator/mod.rs
```diff
@@ -356,24 +359,24 @@ impl ValidatorStore {
...
                 if local_position <= position.delivery_position.position {
                     continue;
                 }
+
                 self.fullnode_store
                     .perpetual_tables
                     .streams
-                    .prefix_iter_at(&subnet, &position)?
+                    .prefix_iter(&(&subnet, &position.delivery_position.position))?
+                    .skip(1)
```
dvdplm, line 371: Is this the "hack" we need to fix the bug? Have you figured out exactly why rocksdb behaves weirdly with `prefix_iter_at`? At Parity many years back we had issues with rocksdb iterators behaving in unexpected ways but I have forgotten the details tbh.

crates/topos-tce-synchronizer/src/lib.rs
```diff
@@ -143,6 +144,12 @@ impl GrpcSynchronizerService for SynchronizerService {
         request: Request<CheckpointRequest>,
     ) -> Result<Response<CheckpointResponse>, Status> {
         let request = request.into_inner();
```
dvdplm, line 146: I have a dumb question: what actually *is* a checkpoint? an you point me to the docs where it is described? The type is a vector of delivery proofs, which is what confuses me a bit I think.

crates/topos-tce-synchronizer/src/lib.rs
```diff
@@ -151,41 +158,62 @@ impl GrpcSynchronizerService for SynchronizerService {
...
+                                    .into_iter()
+                                    .map(|(ready, signature)| SignedReady { ready, signature })
+                                    .collect(),
+                                threshold: v.threshold,
+                            })
+                            .collect();
+                        CheckpointMapFieldEntry {
+                            key: key.to_string(),
+                            value: v,
+                        }
```
dvdplm, line 196: This looks like we're cloning the `ProofOfDelivery`s from the `diff` in order to create the `CheckpointMapFieldEntry`. Can't we just use `clone()`? I must be missing something. Is it the `source_subnet_id` and `delivery_position` the must be plucked out individually?

crates/topos-tce-synchronizer/src/checkpoints_collector/mod.rs
```diff
@@ -184,11 +189,14 @@ impl CheckpointSynchronizer {
...
+        for (subnet, proofs) in diff {
             let len = proofs.len();
             let unverified_certs = self.store.insert_unverified_proofs(proofs)?;
 
-            debug!("Persist {} unverified proofs", len);
+            debug!(
+                "Persist {} unverified proof of delivery for {}",
+                len, subnet
+            );
             certs.extend(&unverified_certs[..]);
```
dvdplm, line 200:
```suggestion
            debug!("Persist {} unverified proofs for {}", proofs.len(), subnet);
```
Doesn't seemt to be used elesewhere?

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
```diff
@@ -230,13 +230,20 @@ impl BroadcastState {
...
             .subscriptions_view
             .network_size
             .checked_sub(self.subscriptions_view.ready.len())
         {
             Some(consumed) => consumed >= self.delivery_threshold,
             None => false,
-        }
+        };
+
+        debug!(
```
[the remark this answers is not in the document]
Freyskeyd: If you consider this particular log as high traffic, I'm suggesting that we create a task to list more logs that are really more critical than this one.

crates/topos-tce-storage/src/tests/checkpoints.rs
Freyskeyd: In this test, the `get_checkpoint_diff` method has an input of `&[checkpoint]` (line 100) while the `get_checkpoint_diff_with_no_input` `get_checkpoint_diff` method doesn't have one (line 57).

crates/topos-tce-storage/src/tests/checkpoints.rs
Freyskeyd: Feel free to create a ticket to add them, as this PR is here to unlock the `testnet` I will not add more things to it.

crates/topos-tce-storage/src/validator/mod.rs
```diff
@@ -336,14 +336,17 @@ impl ValidatorStore {
...
-            .map(|v| (v.delivery_position.subnet_id, vec![v]))
+        let from_positions: HashMap<SubnetId, &ProofOfDelivery> = from
+            .iter()
+            .map(|v| (v.delivery_position.subnet_id, v))
             .collect();
 
+        let mut output: HashMap<SubnetId, Vec<ProofOfDelivery>> =
+            from_positions.iter().map(|(s, _)| (*s, vec![])).collect();
+
         // Request the local head checkpoint
```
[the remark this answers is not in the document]
Freyskeyd: It means the `head` of every source subnet `streams`, a `checkpoint` is a collection of `subnet's position`

crates/topos-tce-storage/src/validator/mod.rs
Freyskeyd: It is due to how we insert the prefix (`bincode` encoded fixed size int of subnet + position) and how we seek the key.

crates/topos-tce-synchronizer/src/checkpoints_collector/mod.rs
Freyskeyd: I don't understand the suggestion

crates/topos-tce-synchronizer/src/lib.rs
Freyskeyd: The related RFC is [here](https://github.com/toposware/rfcs/blob/main/rfcs/0003-local-storage.md)

crates/topos-tce-synchronizer/src/lib.rs
Freyskeyd: This part is building the gRPC response based on the result of the query on the `store`.

crates/topos-tce-synchronizer/src/checkpoints_collector/mod.rs
dvdplm: The suggestion is off, not sure what happened here, maybe some bug in the vscode PR review plugin. Anyway, the suggestion here was to remove the `let len = proofs.len()` line and just use `proofs.len()` in the debug statement. Just a nitpick really.

PR #448, opened 2024-02-06
feat: add benchmark dns option to spam subcommand

gruberb:
# Description

When operating `topos` nodes in a Kubernetes environment, we have to address each service through a Kubernetes abstraction, called a `pod`. These pods are networked via `DNS`. When testing the network via `topos regtest spam` our fixed `IP_ADDRESS:PORT` approach of talking to each node doesn't work. We therefore add new flags to the command:

```
topos regtest spam --benchmark --dns "http://validator-{N}:1340" --number 10
```

Fixes [TP-852](https://linear.app/toposware/issue/TP-852/adjust-the-topos-regtest-spam-command-to-take-dns-instead-of-ip)

## PR Checklist:

### DIFF
--- a/crates/topos-certificate-spammer/src/error.rs
+++ b/crates/topos-certificate-spammer/src/error.rs
@@ -16,6 +16,6 @@ pub enum Error {
     TCENodeConnection(topos_tce_proxy::Error),
     #[error("Certificate signing error: {0}")]
     CertificateSigning(topos_core::uci::Error),
-    #[error("Kubernetes config error: {0}")]
-    KubernetesConfigError(String),
+    #[error("BenchmkarkConfigError config error: {0}")]
+    BenchmarkConfigError(String),
 }
--- a/crates/topos-certificate-spammer/src/error.rs
+++ b/crates/topos-certificate-spammer/src/error.rs
@@ -16,4 +16,6 @@ pub enum Error {
     TCENodeConnection(topos_tce_proxy::Error),
     #[error("Certificate signing error: {0}")]
     CertificateSigning(topos_core::uci::Error),
+    #[error("Kubernetes config error: {0}")]
+    KubernetesConfigError(String),
 }
--- a/crates/topos-certificate-spammer/src/error.rs
+++ b/crates/topos-certificate-spammer/src/error.rs
@@ -17,5 +17,5 @@ pub enum Error {
     #[error("Certificate signing error: {0}")]
     CertificateSigning(topos_core::uci::Error),
     #[error("BenchmkarkConfigError config error: {0}")]
-    BenchmarkConfigError(String),
+    BenchmarkConfig(String),
 }
--- a/crates/topos-certificate-spammer/src/lib.rs
+++ b/crates/topos-certificate-spammer/src/lib.rs
@@ -15,7 +15,7 @@ use tracing::{debug, error, info, info_span, Instrument, Span};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
 mod config;
-mod error;
+pub mod error;
 mod utils;
 
 use error::Error;
@@ -195,11 +195,7 @@ pub async fn run(
     args: CertificateSpammerConfig,
     mut shutdown: mpsc::Receiver<oneshot::Sender<()>>,
 ) -> Result<(), Error> {
-    info!(
-        "Starting topos certificate spammer with the following arguments: {:#?}",
-        args
-    );
-
+    let opts = args.clone();
     // Is list of nodes is specified in the command line use them otherwise use
     // config file provided nodes
     let target_nodes = if args.benchmark {
@@ -216,7 +212,7 @@ pub async fn run(
                 .collect::<Vec<String>>()
         } else {
             return Err(Error::BenchmarkConfig(
-                "hosts pattern or number of nodes not specified".into(),
+                "The --benchmark flag needs two additional flags being passed to it:\n--hosts http://validator-{N}\n--number 10".into(),
             ));
         }
     } else if let Some(nodes) = args.target_nodes {
@@ -278,6 +274,11 @@ pub async fn run(
     let number_of_peer_nodes = target_nodes.len();
     let mut batch_interval = time::interval(Duration::from_millis(args.batch_interval));
     let mut batch_number: u64 = 0;
+
+    info!(
+        "Starting topos certificate spammer with the following arguments: {:#?}",
+        opts
+    );
 
     let shutdown_sender = loop {
         let should_send_batch = tokio::select! {
--- a/crates/topos-certificate-spammer/src/lib.rs
+++ b/crates/topos-certificate-spammer/src/lib.rs
@@ -200,7 +200,12 @@ pub async fn run(
     // config file provided nodes
     let target_nodes = if args.benchmark {
         if let (Some(hosts), Some(number)) = (args.hosts, args.number) {
-            if hosts.replace("{N}", &0.to_string()).parse::<Uri>().is_err() {
+            let uri = hosts
+                .replace("{N}", &0.to_string())
+                .parse::<Uri>()
+                .map_err(|e| Error::BenchmarkConfig(e.to_string()))?;
+
+            if uri.host().is_none() || uri.path().is_empty() || uri.port_u16().is_none() {
                 return Err(Error::BenchmarkConfig(
                     "Invalid hosts pattern. Has to be in the format of http://validator-1:9090"
                         .into(),
@@ -212,7 +217,7 @@ pub async fn run(
                 .collect::<Vec<String>>()
         } else {
             return Err(Error::BenchmarkConfig(
-                "The --benchmark flag needs two additional flags being passed to it:\n--hosts http://validator-{N}\n--number 10".into(),
+                "The --benchmark flag needs the following two additional flags being passed to it:\n--hosts http://validator-{N}\n--number 10".into(),
             ));
         }
     } else if let Some(nodes) = args.target_nodes {
--- a/crates/topos-certificate-spammer/src/lib.rs
+++ b/crates/topos-certificate-spammer/src/lib.rs
@@ -1,5 +1,6 @@
 //! Utility to spam dummy certificates
 
+use http::Uri;
 use opentelemetry::trace::FutureExt;
 use serde::Deserialize;
 use std::collections::HashMap;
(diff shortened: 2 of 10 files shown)


crates/topos-certificate-spammer/src/error.rs
```diff
@@ -16,4 +16,6 @@ pub enum Error {
     TCENodeConnection(topos_tce_proxy::Error),
     #[error("Certificate signing error: {0}")]
     CertificateSigning(topos_core::uci::Error),
+    #[error("Kubernetes config error: {0}")]
```
Freyskeyd, line 19: I feel that it can be misleading, we're not dealing with `kubernetes` config at all here.
Maybe we should rename the `--kubernetes` into `--benchmark` or something similar?

crates/topos-certificate-spammer/src/lib.rs
```diff
@@ -201,7 +201,18 @@ pub async fn run(
 
     // Is list of nodes is specified in the command line use them otherwise use
     // config file provided nodes
-    let target_nodes = if let Some(nodes) = args.target_nodes {
+    let target_nodes = if args.kubernetes {
+        // Kubernetes mode
+        if let (Some(dns), Some(number)) = (args.dns, args.number) {
+            (1..=number)
```
Freyskeyd, line 207: @foo0x29a Does the node will start from `1` or `0` ?

crates/topos-certificate-spammer/src/lib.rs
```diff
@@ -201,7 +201,18 @@ pub async fn run(
...
+                .map(|n| dns.replace("{N}", &n.to_string()))
+                .collect::<Vec<String>>()
```
Freyskeyd, line 209: You can validate that the url is valid using [http::uri](https://docs.rs/http/1.0.0/http/uri/index.html) (and importing it in `Cargo.toml` `http.workspace = true` )

crates/topos-certificate-spammer/src/lib.rs
foo0x29a: Indexes will start from `0`!

crates/topos-certificate-spammer/src/lib.rs
gruberb: Added a check. Debated how exactly we we want to check for a valid `Url`. So I think just doing it once in the beginning is the cheapest, cleanest way.

crates/topos/tests/snapshots/regtest__regtest_spam_help_display.snap
```diff
@@ -33,6 +33,12 @@ Options:
           Socket of the opentelemetry agent endpoint. If not provided open telemetry will not be used [env: TOPOS_OTLP_AGENT=]
       --otlp-service-name <OTLP_SERVICE_NAME>
           Otlp service name. If not provided open telemetry will not be used [env: TOPOS_OTLP_SERVICE_NAME=]
+      --benchmark
```
dvdplm, line 36: In the PR description I read `--kubernetes` but it is actually `--benchmark` right? I mean this here is correct and the PR description is wrong?

crates/topos/tests/snapshots/regtest__regtest_spam_help_display.snap
```diff
@@ -33,6 +33,12 @@ Options:
...
+          Flag to indicate usage of Kubernetes [env: TOPOS_NETWORK_SPAMMER_BENCHMARK=]
+      --dns <DNS>
+          DNS template for generating target node addresses. e.g. `--dns="http://validator-{N}:1340"` [env: TOPOS_NETWORK_SPAMMER_DNS=]
+      --number <NUMBER>
+          Number of nodes to generate based on the DNS template [env: TOPOS_NETWORK_SPAMMER_NUMBER=]
```
dvdplm, line 41: So using `--number` without both `--dns` and `--benchmark` would be an error? I'm a bit concerned about having interdependent switches like that. Can get messy.

crates/topos/tests/snapshots/regtest__regtest_spam_help_display.snap
```diff
@@ -33,6 +33,12 @@ Options:
           Socket of the opentelemetry agent endpoint. If not provided open telemetry will not be used [env: TOPOS_OTLP_AGENT=]
       --otlp-service-name <OTLP_SERVICE_NAME>
           Otlp service name. If not provided open telemetry will not be used [env: TOPOS_OTLP_SERVICE_NAME=]
+      --benchmark
+          Flag to indicate usage of Kubernetes [env: TOPOS_NETWORK_SPAMMER_BENCHMARK=]
+      --dns <DNS>
+          DNS template for generating target node addresses. e.g. `--dns="http://validator-{N}:1340"` [env: TOPOS_NETWORK_SPAMMER_DNS=]
```
dvdplm, line 39: Hmm, what about calling this `--hosts <HOST_TMPL>` instead? The fact that kubernetes wants host names and uses DNS to find peers is sort of a detail that isn't too relevant I think?

crates/topos/tests/snapshots/regtest__regtest_spam_help_display.snap
dvdplm: Is the spammer actually functional without this flag? Not sure what the use case for the cert spammer outside of the kubernetes stuff we're building is. If there is none, then I think it can be the default and save us some API surface to care about.

gruberb: Yeah that's a good point. I was under the assumption that we are still using `regtest spam` in the `local-erc20-infra` for our `e2e` tests. But it doesn't seem like the case.

Technically we can run `regtest spam` in a Docker environment with fixed `IP`s as well, and I wouldn't want to remove that feature. 

WDYT, @Freyskeyd?

crates/topos/tests/snapshots/regtest__regtest_spam_help_display.snap
gruberb: `clap` can `require` flags if another flag is passed. However, we are using `#[derive]` and it doesn't work (yet?) there. I don't know how else to solve this but just by adding maybe a help text?

crates/topos/tests/snapshots/regtest__regtest_spam_help_display.snap
```diff
@@ -33,6 +33,12 @@ Options:
           Socket of the opentelemetry agent endpoint. If not provided open telemetry will not be used [env: TOPOS_OTLP_AGENT=]
       --otlp-service-name <OTLP_SERVICE_NAME>
           Otlp service name. If not provided open telemetry will not be used [env: TOPOS_OTLP_SERVICE_NAME=]
+      --benchmark
+          Flag to indicate usage of Kubernetes [env: TOPOS_NETWORK_SPAMMER_BENCHMARK=]
+      --hosts <HOSTS>
+          DNS template for generating target node addresses. e.g. `--hosts="http://validator-{N}:1340"` [env: TOPOS_NETWORK_SPAMMER_HOSTS=]
```
dvdplm, line 39:
```suggestion
         Template for generating target node hostnames e.g. `--hosts="http://validator-{N}:1340"` [env: TOPOS_NETWORK_SPAMMER_HOSTS=]
```

crates/topos/tests/snapshots/regtest__regtest_spam_help_display.snap
Freyskeyd: We need to keep the possibility to run it locally I guess

crates/topos-certificate-spammer/src/lib.rs
```diff
@@ -194,14 +195,32 @@ pub async fn run(
...
+            }
+
+            (0..number)
+                .map(|n| hosts.replace("{N}", &n.to_string()))
+                .collect::<Vec<String>>()
+        } else {
+            return Err(Error::BenchmarkConfig(
+                "The --benchmark flag needs the following two additional flags being passed to it:\n--hosts http://validator-{N}\n--number 10".into(),
+            ));
+        }
```
Freyskeyd, line 222: I don't think you need all those code as you have the `requires` on the `clap` struct

crates/topos/src/components/regtest/commands/spam.rs
```diff
@@ -67,6 +67,21 @@ pub struct Spam {
...
     pub otlp_service_name: Option<String>,
+    /// Flag to indicate usage of Kubernetes.
+    #[arg(
+        long,
+        env = "TOPOS_NETWORK_SPAMMER_BENCHMARK",
+        requires = "hosts",
+        requires = "number"
+    )]
+    pub benchmark: bool,
+    /// Template for generating target node entrypoints.
```
Freyskeyd, line 78: maybe the variable name should also be `entrypoints` instead of `hosts`

Freyskeyd (approved):
little comments but not blocking

crates/topos/src/components/regtest/commands/spam.rs
dvdplm: How about `target_hosts`?

crates/topos/src/components/regtest/commands/spam.rs
```diff
@@ -67,6 +67,21 @@ pub struct Spam {
...
+    /// e.g. `--hosts="http://validator-{N}:1340"`
```
atanmarko, line 79: What about testing it locally, what pattern should be applied? Add a comment for that

crates/topos-certificate-spammer/src/lib.rs
gruberb: But the Rust compiler doesn't know or validate the `requires` flag. We can do an `unwrap()` to get `args.number` for example. 

I would leave this here, because there is no compiler guarantee us removing the `requires` flag at some point.

crates/topos-certificate-spammer/src/lib.rs
```diff
@@ -261,6 +280,11 @@ pub async fn run(
     let mut batch_interval = time::interval(Duration::from_millis(args.batch_interval));
     let mut batch_number: u64 = 0;
 
+    info!(
+        "Starting topos certificate spammer with the following arguments: {:#?}",
+        opts
```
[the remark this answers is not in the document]
gruberb: Remove it for now. We have other `info` statements. We see if we need more granular `debug` or `info` once we are running the benchmarks.

crates/topos/src/components/regtest/commands/spam.rs
gruberb: The intent is not to run this locally but in a `kubernetes` cluster for now. If ran locally in another `docker` environment, we have the `target-nodes` flag.

crates/topos/src/components/regtest/commands/spam.rs
gruberb: I'll go with `target_hosts`

PR #449, opened 2024-02-06
feat: remove dht publication

Freyskeyd:
# Description

This PR removes the DHT publication of the addresses of the node on the Kademlia DHT.
It is not needed anymore as we rely on `gossipsub` for the broadcast and on `gRPC over p2p` for anything else.

## PR Checklist:


crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -38,32 +27,20 @@ pub struct Runtime {
 
     /// Shutdown signal receiver from the client
     pub(crate) shutdown: mpsc::Receiver<oneshot::Sender<()>>,
+
+    pub(crate) current_bootstrap_id: Option<QueryId>,
```
gruberb, line 31: Could we get a doc comment here?

gruberb (changes requested):
Would be nice to comment the `current_bootstrap_id`. Question also: Why `current_*`? And the type is `QueryId`. Should we rename it then to `query_id`?

crates/topos-p2p/src/runtime/mod.rs
Freyskeyd: the `current_bootstrap_id` will be use to track the different query of the current `bootstrap` of kademlia that we'll perform over time.

dvdplm: @Freyskeyd Can you elaborate on that? I don't quite understand. And +1 for a doc comment here.

docs/architecture/addr_announcer.md
```diff
@@ -1,44 +0,0 @@
-# AddrAnnouncer
```
dvdplm, line 1: Now that this is removed, can we get a description of how this works after the PR?

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -38,32 +27,20 @@ pub struct Runtime {
...
-    pub async fn bootstrap(mut self) -> Result<Self, Box<dyn std::error::Error>> {
-        if self.bootstrapped {
-            return Err(Box::new(P2PError::BootstrapError(
-                "Network is already bootstrapped or in bootstrap",
-            )));
-        }
-
-        self.bootstrapped = true;
-
+    pub async fn bootstrap(&mut self) -> Result<(), Box<dyn std::error::Error>> {
```
dvdplm, line 38: Missing docs. Maybe link to https://docs.rs/libp2p-kad/latest/libp2p_kad/struct.Behaviour.html#method.bootstrap here? Seems like that provides a good description of the bootstrapping process and explains why things are named like they are?

crates/topos/src/components/node/services/process.rs
```diff
@@ -58,7 +58,8 @@ pub(crate) fn spawn_tce_process(
     shutdown: (CancellationToken, mpsc::Sender<()>),
 ) -> JoinHandle<Result<ExitStatus, Errors>> {
     config.boot_peers = genesis
-        .boot_peers(Some(topos_p2p::constants::TCE_BOOTNODE_PORT))
+        // .boot_peers(Some(topos_p2p::constants::TCE_BOOTNODE_PORT))
```
dvdplm, line 61: Leftover?

crates/topos-tce/src/lib.rs
```diff
@@ -104,52 +128,47 @@ pub async fn run(
...
-                validator_store: validator_store.clone(),
-            }),
-        ),
-    );
-
     let certificates_synced = fullnode_store
         .count_certificates_delivered()
         .map_err(|error| format!("Unable to count certificates delivered: {error}"))
-        .unwrap();
+        .expect("Unable to count certificates delivered");
```
gruberb, line 134: Instead of `.expect`, we can just use `?`. We repeat the same `error` here. So we either remove `map_err`, or use `?` instead of `expect`.

Same for `139` and `144`.

gruberb (approved):

crates/topos-p2p/src/behaviour/discovery.rs
```diff
@@ -48,10 +59,41 @@ impl DiscoveryBehaviour {
...
+            inner: kademlia,
+            current_bootstrap_query_id: None,
+            next_bootstrap_query: Some(Box::pin(tokio::time::interval(config.bootstrap_interval))),
+        }
+    }
+
+    /// Start the kademlia bootstrap process if it is not already in progress.
+    /// The bootstrap process is used to discover new peers in the network.
+    /// Starting by sending a `FIND_NODE` query of the local PeerId in the DHT.
+    /// Then random PeerId are forged in order to randomly walk the network.
```
dvdplm, line 79:
```suggestion
    /// Then a random PeerId is created in order to randomly walk the network.
```

crates/topos-p2p/src/behaviour/discovery.rs
```diff
@@ -48,10 +59,41 @@ impl DiscoveryBehaviour {
...
+        Self {
+            inner: kademlia,
+            current_bootstrap_query_id: None,
+            next_bootstrap_query: Some(Box::pin(tokio::time::interval(config.bootstrap_interval))),
+        }
+    }
+
+    /// Start the kademlia bootstrap process if it is not already in progress.
+    /// The bootstrap process is used to discover new peers in the network.
+    /// Starting by sending a `FIND_NODE` query of the local PeerId in the DHT.
```
dvdplm, line 78:
```suggestion
    /// The bootstrap process starts by sending a `FIND_NODE` query for the local PeerId in the DHT.
```

dvdplm (approved):

PR #451, opened 2024-02-13
chore: increase payload size for sync in config

Freyskeyd:
# Description

This PR backport fixes from another PR related to sync and storage.
It also adds a new configuration for the sync:

```
[tce.synchronization]
interval-seconds = 10
limit-per-subnet = 100
```

## PR Checklist:


crates/topos-tce-synchronizer/src/lib.rs
```diff
@@ -150,6 +151,14 @@ impl GrpcSynchronizerService for SynchronizerService {
             .unwrap_or(Uuid::new_v4());
         debug!("Received request for checkpoint (request_id: {})", id);
 
+        let limit_per_subnet: usize = max(
```
Freyskeyd, line 154: I put `max` for backward compatibility on gRPC, if we receive a message without `limit_per_subnet` it gonna be set to `0`.

crates/topos-core/proto/topos/tce/v1/synchronization.proto
```diff
@@ -18,6 +18,8 @@ message CheckpointRequest {
   topos.shared.v1.UUID request_id = 1;
 
   repeated ProofOfDelivery checkpoint = 2;
+
+  uint64 limit_per_subnet = 3;
```
dvdplm, line 22: This is the state sync batch size (per subnet) right?

dvdplm (approved):
lgtm. Can we also increase the max GRPC payload from 4 to 10? And where is the configuration file we use in testnet, in the ansible playbook perhaps?

PR #454, opened 2024-02-14
feat: track queue size in metrics/graphql

Freyskeyd:
# Description

This PR adds two new metrics accessible on the default port `3000` (e.g.: `localhost:3000`):

- `STORAGE_PENDING_POOL_COUNT` that tracks the number of certificates in the `pending_pool`.
- `STORAGE_PRECEDENCE_POOL_COUNT` that tracks the number of certificates in the `precedence_pool`.

:warning: Keep in my that those values are estimation based on the `rocksdb` metadata:

> Q: Why GetIntProperty can only returns an estimated number of keys in a RocksDB database?
>
> A: Obtaining an accurate number of keys in any LSM databases like RocksDB is a challenging problem as they have duplicate keys and deletion entries (i.e., tombstones) that will require a full compaction in order to get an accurate number of keys. In addition, if the RocksDB database contains merge operators, it will also make the estimated number of keys less accurate.

The `GraphQL` API is exposing those value under `getStoragePoolStats`.

## PR Checklist:

### DIFF
--- a/crates/topos-tce-api/tests/runtime.rs
+++ b/crates/topos-tce-api/tests/runtime.rs
@@ -21,6 +21,7 @@ use topos_core::{
     },
     uci::Certificate,
 };
+use topos_metrics::{STORAGE_PENDING_POOL_COUNT, STORAGE_PRECEDENCE_POOL_COUNT};
 use topos_tce_api::{Runtime, RuntimeEvent};
 use topos_tce_storage::types::CertificateDeliveredWithPositions;
 use topos_tce_storage::StorageClient;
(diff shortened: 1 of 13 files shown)


crates/topos-tce-api/tests/runtime.rs
```diff
@@ -621,3 +622,73 @@ async fn can_query_graphql_endpoint_for_certificates(
...
+        .await;
+
+    // Wait for server to boot
+    tokio::time::sleep(Duration::from_millis(100)).await;
+
+    let query = "query {getStoragePoolStats}";
+
+    #[derive(Debug, Deserialize)]
+    struct Response {
+        // data: HashMap<String, serde_json::Value>,
```
dvdplm, line 662: Can remove I think.

crates/topos-tce-storage/src/fullnode/mod.rs
```diff
@@ -232,8 +233,11 @@ impl WriteStore for FullNodeStore {
 }
 
 impl ReadStore for FullNodeStore {
-    fn count_certificates_delivered(&self) -> Result<usize, StorageError> {
-        Ok(self.perpetual_tables.certificates.iter()?.count())
+    fn count_certificates_delivered(&self) -> Result<u64, StorageError> {
```
dvdplm, line 236: Maybe we should call this `estimate_certs_delivered`?

crates/topos-tce-storage/src/errors.rs
```diff
@@ -33,6 +33,9 @@ pub enum InternalStorageError {
     #[error("Invalid query argument: {0}")]
     InvalidQueryArgument(&'static str),
 
+    #[error("Unexpected DB state: {0}")]
+    UnexpectedDBState(&'static str),
```
dvdplm, line 37: It's not really a problem with the DB *state* if a property is missing. Wouldn't this error be for a situation where there's a new/old/incompatible rocksdb version that does not support some property?

dvdplm (approved):
LGTM. All my comments are nits that can wait.

PR #455, opened 2024-02-14
fix: add next_pending_certificate on end task

Freyskeyd:
# Description

This PR adds a call to `get_next_pending_certificates` after a task ends in order to process next pending certificates.
It also removes some old struct/mechanism of the broadcast.

## PR Checklist:

### DIFF
--- a/crates/topos-tce-broadcast/benches/task_manager.rs
+++ b/crates/topos-tce-broadcast/benches/task_manager.rs
@@ -56,7 +56,7 @@ pub async fn processing_double_echo(n: u64, validator_store: Arc<ValidatorStore>
         cmd_receiver,
         event_sender,
         double_echo_shutdown_receiver,
-        validator_store,
+        validator_store.clone(),
         broadcast_sender,
     );
 
@@ -82,7 +82,9 @@ pub async fn processing_double_echo(n: u64, validator_store: Arc<ValidatorStore>
         .collect::<Vec<_>>();
 
     for cert in &certificates {
-        double_echo.broadcast(cert.certificate.clone(), true).await;
+        _ = validator_store
+            .insert_pending_certificate(&cert.certificate)
+            .unwrap();
     }
 
     for cert in &certificates {
--- a/crates/topos-tce-broadcast/src/task_manager/mod.rs
+++ b/crates/topos-tce-broadcast/src/task_manager/mod.rs
@@ -87,6 +87,9 @@ impl TaskManager {
         }
     }
 
+    /// Fetch the next pending certificates from the storage and create tasks for them.
+    /// This method is called periodically to check for new pending certificates and when
+    /// a task has finished.
     fn next_pending_certificate(&mut self) {
         debug!("Checking for next pending_certificates");
         match self.validator_store.get_next_pending_certificates(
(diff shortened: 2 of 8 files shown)


crates/topos-tce-broadcast/src/task_manager/mod.rs
```diff
@@ -136,9 +146,12 @@ impl TaskManager {
                         debug!("Task for certificate {} finished successfully", certificate_id);
                         self.tasks.remove(&certificate_id);
                         DOUBLE_ECHO_ACTIVE_TASKS_COUNT.dec();
+
                     } else {
                         debug!("Task for certificate {} finished unsuccessfully", certificate_id);
                     }
+
+                    self.next_pending_certificate();
```
dvdplm, line 154: So this is the key addition in the PR? What is the thinking here?

crates/topos-tce-broadcast/src/task_manager/mod.rs
```diff
@@ -87,6 +87,29 @@ impl TaskManager {
         }
     }
 
+    fn next_pending_certificate(&mut self) {
+        debug!("Checking for next pending_certificates");
+        match self.validator_store.get_next_pending_certificates(
+            &self.latest_pending_id,
+            *PENDING_LIMIT_PER_REQUEST_TO_STORAGE,
```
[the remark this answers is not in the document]
Freyskeyd: Yes it by default to `1000` and can be change using env var `TOPOS_PENDING_LIMIT_PER_REQUEST_TO_STORAGE` for debug or development use. If we want to provide that kind of configuration for the admin we need to expose them as config file params.

crates/topos-tce-broadcast/src/task_manager/mod.rs
Freyskeyd: When a `task` finish, we ask for more work instead of waiting the next tick.

crates/topos-tce-broadcast/benches/task_manager.rs
```diff
@@ -82,7 +82,9 @@ pub async fn processing_double_echo(n: u64, validator_store: Arc<ValidatorStore>
         .collect::<Vec<_>>();
 
     for cert in &certificates {
-        double_echo.broadcast(cert.certificate.clone(), true).await;
+        _ = validator_store
+            .insert_pending_certificate(&cert.certificate)
+            .unwrap();
```
gruberb, line 87: Let's add error handling here. What does it mean if we cannot add it to the store? What should we do then?

gruberb (changes requested):
Remove `.unwrap()`

crates/topos-tce-broadcast/benches/task_manager.rs
Freyskeyd: as it is for `cargo benchmark` that we do not currently use I was thinking it was ok to panic with the `StorageError` directly instead of creating a complex error handling.

crates/topos-tce-broadcast/src/task_manager/mod.rs
dvdplm: Do you think we need to look into having shorter ticks, the default 100ms is quite long and afaik we have CPU capacity to spare in production? (Not for this PR of course, but it could be useful to expose in the config).

dvdplm (approved):

PR #456, opened 2024-02-15
chore: increase payload size for sync in config (#451)

Freyskeyd:
# Description

Adds a manual compaction when opening the DB on `pending_pool` and `precedence_pool` CF
## PR Checklist:


dvdplm (approved):
I'm ok doing this, but there's a risk here: on a big rocksdb, doing compaction can take a long time and afaik there is no way to know just how long it takes.

We can mitigate this risk by making it optional by way of a config option, or have a subcommand that does the compaction while the node is running (this is safe because rocksdb runs compactions on its own whenever it wants to).

PR #457, opened 2024-02-16
Rename Sequencer –> Certificate Producer

dvdplm:
Renaming the Sequencer to Certificate Producer in the whole repo.


atanmarko:
How this happened?

atanmarko (approved):

atanmarko:
@dvdplm I would prefer `cert-producer` instead of `certificate-producer`. It looks a bit unwieldy in the crate name `topos-certificate-producer-subnet-runtime`

dvdplm:
> I would prefer `cert-producer` instead of `certificate-producer`. It looks a bit unwieldy in the crate name `topos-certificate-producer-subnet-runtime`

I agree that the name is unwieldy. I would not mind settling on `cert-producer`. @hadjiszs What do you say?

PR #458, opened 2024-02-16
fix: remove duplicated certificate push on gossipsub

hadjiszs:
# Description

The initial sender is supposed to be the only one to publish the Certificate on the gossipsub topic `topos_gossip`. The initial sender is the one who receives the Certificate from the gRPC API and not from gossipsub.

The rework on the storage by moving the management of the pending and precedence pool on the storage layer introduced the loss of the origin of the Certificate (whether it is received from gossipsub or grpc), consequently, the message was pushed by everyone on the `topos_gossip` topic on both cases since this storage rework.

## Additions and Changes

- Before (default config, see [here](https://docs.rs/libp2p-gossipsub/latest/libp2p_gossipsub/struct.Config.html#method.message_id))
  - Author based message id: `(PeerId, nonce)`, so we had as many duplicate of Certificate as the number of peers
- After
  - Content based message id: `hash(payload)`, consequently the other peers will detect the payload being the same, so it will not be published nor downloaded multiple times

## PR Checklist:

### DIFF
--- a/crates/topos-p2p/src/behaviour/gossip.rs
+++ b/crates/topos-p2p/src/behaviour/gossip.rs
@@ -76,7 +76,7 @@ impl Behaviour {
                 // Content based id
                 let mut s = DefaultHasher::new();
                 msg_id.data.hash(&mut s);
-                gossipsub::MessageId::from(s.finish().to_string())
+                gossipsub::MessageId::from(s.finish().to_be_bytes())
             })
             .build()
             .unwrap();
--- a/crates/topos-p2p/src/behaviour/gossip.rs
+++ b/crates/topos-p2p/src/behaviour/gossip.rs
@@ -1,3 +1,5 @@
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
 use std::{
     collections::{HashMap, HashSet, VecDeque},
     env,
@@ -35,7 +37,9 @@ impl Behaviour {
     ) -> Result<usize, &'static str> {
         match topic {
             TOPOS_GOSSIP => {
-                _ = self.gossipsub.publish(IdentTopic::new(topic), message);
+                if let Ok(msg_id) = self.gossipsub.publish(IdentTopic::new(topic), message) {
+                    debug!("Published on topos_gossip: {:?}", msg_id);
+                }
             }
             TOPOS_ECHO | TOPOS_READY => self.pending.entry(topic).or_default().push_back(message),
             _ => return Err("Invalid topic"),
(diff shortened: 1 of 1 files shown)


crates/topos-p2p/src/behaviour/gossip.rs
```diff
@@ -35,7 +37,9 @@ impl Behaviour {
     ) -> Result<usize, &'static str> {
         match topic {
             TOPOS_GOSSIP => {
-                _ = self.gossipsub.publish(IdentTopic::new(topic), message);
+                if let Ok(msg_id) = self.gossipsub.publish(IdentTopic::new(topic), message) {
```
Freyskeyd, line 40: If we receive a `Gossip` message at minute `M` and put it in the `precedence_pool` and start delivering it at `M+2`, the `publish` method will republish it on the network. The default configuration of the key tracking is 1 min TTL. I don't think it is an issue, but we need to keep in mind that the message IDs are cached in an LRU, which doesn't last long.

Freyskeyd (approved):

crates/topos-p2p/src/behaviour/gossip.rs
```diff
@@ -68,6 +72,12 @@ impl Behaviour {
         let gossipsub = gossipsub::ConfigBuilder::default()
             .max_transmit_size(2 * 1024 * 1024)
             .validation_mode(gossipsub::ValidationMode::Strict)
+            .message_id_fn(|msg_id| {
+                // Content based id
+                let mut s = DefaultHasher::new();
+                msg_id.data.hash(&mut s);
+                gossipsub::MessageId::from(s.finish().to_string())
```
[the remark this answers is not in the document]
dvdplm: There's also the "fast_message_id` feature but I'm not familiar with the potential problems (if any).

dvdplm (approved):

crates/topos-p2p/src/behaviour/gossip.rs
dvdplm: Ah, this is the `duplicate_cache_time` (https://docs.rs/libp2p/latest/libp2p/gossipsub/struct.Config.html#method.duplicate_cache_time). The size is not bounded, and the TTL is 1min by default.

crates/topos-p2p/src/behaviour/gossip.rs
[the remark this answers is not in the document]
hadjiszs: >should work I think?

seems to work :ok_hand: 

>There's also the "fast_message_id` feature but I'm not familiar with the potential problems (if any).

not sure either

crates/topos-p2p/src/behaviour/gossip.rs
Freyskeyd: Yes, the `duplicate_cache` LRU is used in to define if a message should be sent to the network. It can be changed but the idea remains, some certificate could be gossip more than once.

hadjiszs: Ok I didn't have that in mind

Still better than before where we could potentially have had all peers pushing their instance of the payload more than once, while now it would only be the initial publisher attempting to publish it again, right?

PR #459, opened 2024-02-16
feat: introduce topos-node crate

Freyskeyd:
# Description

This PR introduces a dedicated crate for the `topos` runtime, decoupled from the `topos` CLI that can be used by node operator.
The purpose of the `topos-node` crate is to have everything to run a node from a config file.

This PR is just moving things around and the topos CLI should keep working as expected.
The `topos node up` command is going to be deprecated, the replacement being the `topos-node` binary.

## PR Checklist:


dvdplm:
It would be good to have a sibling PR for the topos-playground as a sort of test that everything works as expected after this lands.

Freyskeyd:
This PR is backward compatible with the current CLI command. I think you mean updating the `local-erc20` right?

dvdplm:
> I think you mean updating the `local-erc20` right?

Yeah, I think that's what I meant! :)

I was thinking about how the topos-playground exercises the whole stack, but you're right, it's the local-erc20 image that is relevant here.

crates/topos-node/Cargo.toml
```diff
@@ -0,0 +1,74 @@
+[package]
+name = "topos-node"
+version = "0.1.0"
+edition = "2021"
+
```
dvdplm, line 5: Can we add the `description` key here too?

crates/topos-node/build.rs
```diff
@@ -0,0 +1,23 @@
+use std::process::Command;
+
+const DEFAULT_VERSION: &str = "detached";
+
+fn main() {
+    // Set TOPOS_VERSION to HEAD short commit hash if None
```
dvdplm, line 6:
```suggestion
    // Set TOPOS_VERSION to HEAD short commit hash unless it's already set
```

dvdplm: Is this env variable typically set to something other than the short hash? What do you think about not making this configurable and just impose the short hash?

crates/topos-node/src/lib.rs
```diff
@@ -0,0 +1,221 @@
+//! Temporary lib exposition for backward topos CLI compatibility
```
dvdplm, line 1: If this is temporary, what is the long term plan?

crates/topos-node/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+    verbose: u8,
+    no_color: bool,
+    otlp_agent: Option<String>,
+    otlp_service_name: Option<String>,
+    no_edge_process: bool,
+    node_path: PathBuf,
+    home: PathBuf,
+    config: NodeConfig,
+) -> Result<(), Box<dyn std::error::Error>> {
+    println!(
```
dvdplm, line 37: Why not `log::info!` this? Doesn't this make it trickier for the tools guys to capture the startup logs? Also: can we print the version here too? (Or are we doing that already somewhere else perhaps?)

crates/topos-node/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+        Ok(genesis) => genesis,
+        Err(_) => {
+            println!(
+                "Could not load genesis.json file on path {} \n Please make sure to have a valid \
+                 genesis.json file for your subnet in the {}/subnet/{} folder.",
+                genesis_file_path.display(),
+                home.display(),
+                &config.base.subnet
+            );
+            std::process::exit(1);
```
dvdplm, line 58: This is great. Should we also `log::error!` this same message?

crates/topos-node/src/main.rs
```diff
@@ -0,0 +1,4 @@
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
dvdplm, line 2: dumb question maybe: why do we need a `main` here?

crates/topos/tests/config.rs
```diff
@@ -15,28 +14,30 @@ use crate::utils::setup_polygon_edge;
...
+
     use super::*;
 
+    #[rstest]
     #[tokio::test]
-    async fn handle_command_init() -> Result<(), Box<dyn std::error::Error>> {
-        let tmp_home_dir = tempdir()?;
-        let path = setup_polygon_edge(tmp_home_dir.path().to_str().unwrap()).await;
+    async fn handle_command_init(create_folder: PathBuf) -> Result<(), Box<dyn std::error::Error>> {
+        let home = create_folder;
```
dvdplm, line 25: Why rebind like this? Why not just rename the `create_folder` param to `home`?

crates/topos/tests/config.rs
```diff
@@ -418,14 +419,27 @@ mod serial_integration {
...
+        if let Ok(output) = cmd.wait_with_output().await {
+            assert!(output.status.success());
+            let stdout = output.unwrap().unwrap().stdout;
+            let stdout = String::from_utf8_lossy(&stdout);
+
+            let reg =
+                Regex::new(r#"Local node is listening on "\/ip4\/.*\/tcp\/9091\/p2p\/"#).unwrap();
+            assert!(reg.is_match(&stdout));
+        } else {
+            panic!("Failed to gracefull shutdown");
```
dvdplm, line 441:
```suggestion
            panic!("Failed to shutdown gracefully");
```

dvdplm (approved):

crates/topos-node/build.rs
Freyskeyd: This could be the `v0.0.XX` version but we didn't update the build process to make it passes through I think

crates/topos-node/src/lib.rs
Freyskeyd: the `tracing` setup was define after, but it is fixed in another PR, this PR is just moving code around.

crates/topos-node/src/lib.rs
Freyskeyd: Maybe if the tracing setup is done, will create a little ticket for that

crates/topos-node/src/main.rs
Freyskeyd: The goal is to be able to launch `topos-node` binary directly

crates/topos-node/src/lib.rs
Freyskeyd: the long term plan is to have the `topos cli` on one side and the `topos-node` binary that represent the whole runtime of a node

crates/topos-node/src/main.rs
dvdplm: So this is a placeholder of sorts then?

crates/topos-node/build.rs
dvdplm: I think having the short hash is so useful so often that it's worth having it always. When we pass through the `x.y.z` string I still think we should include the short hash in the version string, e.g. `x.y.z-abc123abc`

crates/topos-node/src/main.rs
Freyskeyd: Yes we can say that it's a placeholder

dvdplm (approved):

PR #463, opened 2024-02-23
feat: terminate stream if client is dropping the connection

gruberb:
# Description

In our current setup, the `sequencer` is opening a gRPC stream to the `tce`. If successful, we send `delivered_certificates` back to the `sequencer` (or any other client). However, if `sequencer` shuts down (or restarts), the stream was not closed on the `tce` side, resulting in multiple open streams with dangling connections.

This PR is fixing this issue by handling the error cases of a streaming message.

Fixes [TP-500](https://linear.app/toposware/issue/TP-500/as-topos-i-want-to-better-manage-closed-grpc-streams-so-that-i-can)

## Result

```bash
024-02-23T15:55:02.245162Z  INFO topos_tce_api::stream: Received an OpenStream command for the stream cbb40b6a-89fc-497a-8594-fa680656bba1
2024-02-23T15:55:02.245189Z  INFO topos_tce_api::runtime: Stream cbb40b6a-89fc-497a-8594-fa680656bba1 is registered as subscriber

... 

2024-02-23T15:55:37.898530Z ERROR topos_tce_api::stream: Stream error: StreamError { stream_id: cbb40b6a-89fc-497a-8594-fa680656bba1, kind: Transport(Unknown) }
2024-02-23T15:55:37.898598Z  INFO topos_tce_api::runtime: Stream cbb40b6a-89fc-497a-8594-fa680656bba1 terminated gracefully
```

## Left ToDo

- [x] Check if we have to clean up the connection from `active_streams`
- [x] Check if we have to clean up any open channels
- [x] Testing

## PR Checklist:

### DIFF
--- a/crates/topos-tce-api/src/stream/mod.rs
+++ b/crates/topos-tce-api/src/stream/mod.rs
@@ -160,17 +160,16 @@ impl Stream {
                         Err(error) => {
                             match error.kind {
                                 StreamErrorKind::StreamClosed => {
-                                    // Currently, when the client is closing the connection,
-                                    // we don't receive a StreamClosed error. Further investigation is needed.
                                     warn!("Stream {} closed", self.stream_id);
-                                    break;
+                                    return Err(StreamError::new(self.stream_id, StreamErrorKind::StreamClosed));
                                 }
                                 _ => {
                                     // We are not handling specific errors for now.
                                     // If the sequencer is closing the connection, we are receiving a
                                     // StreamErrorKind::TransportError.
                                     error!( "Stream error: {:?}", error);
-                                    break;
+                                    return Err(StreamError::new(self.stream_id, error.kind));
+
                                 }
 
                             }
--- a/crates/topos-tce-api/src/stream/mod.rs
+++ b/crates/topos-tce-api/src/stream/mod.rs
@@ -150,6 +150,8 @@ impl Stream {
                     }
                 }
 
+                // We currently open the stream, but no other message from the client is getting processed.
+                // We are using this open connection to communicate `delivered_certificates` to the client.
                 stream_packet = self.inbound_stream.next() => {
                       if let Some(stream_packet) = stream_packet {
                           match stream_packet {
@@ -157,6 +159,7 @@ impl Stream {
                                     debug!("Received message for stream: {request_id:?}");
                                 }
                                 Err(error) => {
+                                    // In case the stream is getting closed from the client side for example
                                     error!("Stream error: {error:?}");
                                     break;
                                 }
--- a/crates/topos-tce-api/src/stream/mod.rs
+++ b/crates/topos-tce-api/src/stream/mod.rs
@@ -158,9 +158,17 @@ impl Stream {
                             debug!("Received message from stream_id: {:?}", self.stream_id);
                         }
                         Err(error) => {
-                            // In case the stream is getting closed from the client side for example
-                            error!("Stream closed! StreamId: {}", error.stream_id);
-                            break;
+                            match error.kind {
+                                StreamErrorKind::StreamClosed => {
+                                    warn!("Stream {} closed", self.stream_id);
+                                    break;
+                                }
+                                _ => {
+                                    error!( "Stream error: {:?}", error);
+                                    break;
+                                }
+
+                            }
                         }
                     }
                 }
--- a/crates/topos-tce-api/src/stream/mod.rs
+++ b/crates/topos-tce-api/src/stream/mod.rs
@@ -152,25 +152,29 @@ impl Stream {
 
                 // We currently open the stream, but no other message from the client is getting processed.
                 // We are using this open connection to communicate `delivered_certificates` to the client.
-                stream_packet = self.inbound_stream.next() => {
-                      if let Some(stream_packet) = stream_packet {
-                          match stream_packet {
-                                Ok((request_id, _message)) => {
-                                    debug!("Received message for stream: {request_id:?}");
-                                }
-                                Err(error) => {
-                                    // In case the stream is getting closed from the client side for example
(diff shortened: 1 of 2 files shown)


crates/topos-tce-api/src/stream/mod.rs
```diff
@@ -150,14 +150,25 @@ impl Stream {
...
 
-                Some(_stream_packet) = self.inbound_stream.next() => {
-
+                // We currently open the stream, but no other message from the client is getting processed.
+                // We are using this open connection to communicate `delivered_certificates` to the client.
+                Some(stream_packet) = self.inbound_stream.next() => {
+                    match stream_packet {
+                        Ok((request_id, _message)) => {
+                            debug!("Received message for stream: {request_id:?}");
+                        }
```
hadjiszs, line 159: The log message says like we show the stream id no?

crates/topos-tce-api/src/stream/mod.rs
```diff
@@ -155,7 +155,7 @@ impl Stream {
                 Some(stream_packet) = self.inbound_stream.next() => {
                     match stream_packet {
                         Ok((request_id, _message)) => {
-                            debug!("Received message for stream: {request_id:?}");
+                            debug!("Received message from stream_id: {request_id:?}");
```
hadjiszs, line 158: `stream_id: {request_id}`

My previous comment was that we read `... stream: {request_id}`, so I'm confused whether you want to print here `self.stream_id` or `request_id`

gruberb: Ah of course. I got confused. Yes, I wanted to print `self.stream_id` instead of the `request_id`.

crates/topos-tce-api/src/stream/mod.rs
```diff
@@ -150,14 +150,25 @@ impl Stream {
...
                 }
 
-                Some(_stream_packet) = self.inbound_stream.next() => {
-
+                // We currently open the stream, but no other message from the client is getting processed.
+                // We are using this open connection to communicate `delivered_certificates` to the client.
+                Some(stream_packet) = self.inbound_stream.next() => {
+                    match stream_packet {
+                        Ok((_request_id, _message)) => {
+                            debug!("Received message from stream_id: {:?}", self.stream_id);
```
dvdplm, line 158: Should probably be `trace`?

dvdplm: Do we have any good way of correlating a stream_id with the actual nodes involved? If not, is it still useful?

crates/topos-tce-api/src/stream/mod.rs
```diff
@@ -150,14 +150,25 @@ impl Stream {
...
+                        }
+                        Err(error) => {
+                            // In case the stream is getting closed from the client side for example
+                            error!("Stream closed! StreamId: {}", error.stream_id);
+                            break;
```
dvdplm, line 163: Will this trigger a reconnect? Or is the other side expected to reconnect?

Do you think this should be `warn` perhaps? I'm thinking that this condition will happen every time a Certificate Producer is restarting, so it's not *really* an error condition but rather something expected?

crates/topos-tce-api/src/stream/mod.rs
gruberb: I used it to test locally and thought I leave it in here to explain what happens, logically. We don't send any message to the stream from the client and if so, do nothing with it. Maybe removing the `log` all together and leaving a comment is a better choice.

crates/topos-tce-api/src/stream/mod.rs
gruberb: After a round of debugging again, the `error` we receive from a disconnecting client is `TransportError`. To test this I manually stopped the `topos-sequencer`, and therefore the `stream`. I assume a "hard reboot" will lead to this error and not a nice "StreamClosed". 

There are multiple StreamErrors combined in this `error!`, the most likely a hard reboot in our case. So for now, I would leave it at this and investigate in another issue at some point how to we can get a reliable `Error` out of a disconnecting stream.

crates/topos-tce-api/src/stream/mod.rs
dvdplm: A `trace` is also fin for me. Your call!

crates/topos-tce-api/src/stream/mod.rs
```diff
@@ -150,14 +150,38 @@ impl Stream {
...
+                Some(stream_packet) = self.inbound_stream.next() => {
+                    match stream_packet {
+                        Ok((_request_id, _message)) => {
+                            trace!("Received message from stream_id: {:?}", self.stream_id);
+                        }
+                        Err(error) => {
+                            match error.kind {
+                                StreamErrorKind::StreamClosed => {
+                                    // Currently, when the client is closing the connection,
+                                    // we don't receive a StreamClosed error. Further investigation is needed.
```
Freyskeyd, line 164: `StreamClosed` error are generated when the server detect a problem with either the connection or the protocol used and thus close the connection.

Freyskeyd (commented):
Could be good to have tests to cover that a stream close as expected now. I can see that there is two test placeholder that can fit this:

`crates/topos-tce-api/src/stream/tests.rs`

```rust
#[test(tokio::test)]
#[ignore = "not yet implemented"]
async fn closing_client_stream() {}

#[test(tokio::test)]
#[ignore = "not yet implemented"]
async fn closing_server_stream() {}
```

gruberb:
@Freyskeyd can be re-reviewed! Added a test for `closing_client_stream`. Will add `closing_server_stream` in TP-887 where we investigate closing the `tce` side of the stream.

dvdplm (approved):

PR #464, opened 2024-02-27
feat: add p2p layer health check

Freyskeyd:
# Description

This PR is adding and refactoring how the node will setup and validate it's p2p layer connectivity.
It is a first step to allow nodes to be spawn in any order and gain more confidence on the node status.

## Changes

- `discovery` behaviour will periodically walk the network to find new peers and discover more nodes.
	- The default setting is currently fixed to 60s but will be configurable in another PR
	- The first query is targeting one bootnode, if the query fails, the interval between bootstrap query is set to 5s (also configurable later)
- `gossip` behaviour will subscribe to different topics and become healthy when at least one is ok
- the `handle_event` method of the P2P `runtime` is returning a `Result` which allows us to report error (and maybe crash) if needed.
- `StateMachine` in P2P layer represents a struct that hold state projection (to avoid extra calculation), it may be refactored to `enum` and better integrated with the events but for now it is enough.

## Steps

- The P2P layer is starting by `listening` and exposing its addresses.
- Then the `discovery` behaviour will initiate `bootstrap` query targeting the bootnode.
    The connection and result of this query will define if we go or not to the next step.
- The `gossipsub` behaviour will then trigger the `subscribe` to the topics
- When succeed the P2P layer is considered healthy and can continue

## PR Checklist:

### DIFF
--- a/crates/topos-config/src/edge/command.rs
+++ b/crates/topos-config/src/edge/command.rs
@@ -64,6 +64,7 @@ impl CommandConfig {
     }
 
     pub async fn spawn(self) -> Result<ExitStatus, std::io::Error> {
+        info!("Spawning Polygon Edge with args: {:?}", self.binary_path);
         let mut command = Command::new(self.binary_path);
         command.kill_on_drop(true);
         command.args(self.args);
--- a/crates/topos-p2p/src/behaviour/discovery.rs
+++ b/crates/topos-p2p/src/behaviour/discovery.rs
@@ -74,6 +74,9 @@ impl DiscoveryBehaviour {
         Self {
             inner: kademlia,
             current_bootstrap_query_id: None,
+            // If the `discovery` behaviour is created without known_peers
+            // The bootstrap query interval is disabled only when the local
+            // node is a lonely bootnode, other nodes will join it.
             next_bootstrap_query: if known_peers.is_empty() {
                 None
             } else {
--- a/crates/topos-p2p/src/behaviour/discovery.rs
+++ b/crates/topos-p2p/src/behaviour/discovery.rs
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 use std::pin::Pin;
 use std::task::Poll;
+use std::time::Duration;
 
 use crate::error::P2PError;
 use crate::{config::DiscoveryConfig, error::CommandExecutionError};
@@ -18,6 +19,8 @@ use libp2p::{
 use tokio::sync::oneshot;
 use tracing::{debug, error, info};
 
+use super::HealthStatus;
+
 pub type PendingRecordRequest = oneshot::Sender<Result<Vec<Multiaddr>, CommandExecutionError>>;
 
 /// DiscoveryBehaviour is responsible to discover and manage connections with peers
@@ -29,6 +32,8 @@ pub(crate) struct DiscoveryBehaviour {
     pub(crate) current_bootstrap_query_id: Option<QueryId>,
     /// The next bootstrap query interval used to schedule the next bootstrap query
     pub(crate) next_bootstrap_query: Option<Pin<Box<tokio::time::Interval>>>,
+    /// The health status of the discovery behaviour
+    pub(crate) health_status: HealthStatus,
 }
 
 impl DiscoveryBehaviour {
@@ -70,6 +75,7 @@ impl DiscoveryBehaviour {
             inner: kademlia,
             current_bootstrap_query_id: None,
             next_bootstrap_query: Some(Box::pin(tokio::time::interval(config.bootstrap_interval))),
+            health_status: Default::default(),
         }
     }
 
@@ -79,33 +85,21 @@ impl DiscoveryBehaviour {
     /// Then multiple random PeerId are created in order to randomly walk the network.
     pub fn bootstrap(&mut self) -> Result<(), P2PError> {
         if self.current_bootstrap_query_id.is_none() {
-            match self.inner.bootstrap() {
-                Ok(query_id) => {
-                    info!("Started kademlia bootstrap with query_id: {query_id:?}");
-                    self.current_bootstrap_query_id = Some(query_id);
-                }
-                Err(error) => {
-                    error!("Unable to start kademlia bootstrap: {error:?}");
-                    return Err(P2PError::BootstrapError(
-                        "Unable to start kademlia bootstrap",
-                    ));
-                }
-            }
+            let query_id = self.inner.bootstrap()?;
+            debug!("Started kademlia bootstrap query with query_id: {query_id:?}");
+            self.current_bootstrap_query_id = Some(query_id);
         }
 
         Ok(())
     }
 
-    pub fn get_addresses_of_peer(&mut self, peer_id: &PeerId) -> Vec<Multiaddr> {
-        if let Some(key_ref) = self.inner.kbucket(*peer_id) {
-            key_ref
-                .iter()
-                .filter(|e| e.node.key.preimage() == peer_id)
-                .map(|e| e.node.value.first().clone())
-                .collect()
-        } else {
-            Vec::new()
+    /// Change the interval of the next bootstrap queries
+    pub fn change_interval(&mut self, duration: Duration) -> Result<(), P2PError> {
+        if let Some(interval) = self.next_bootstrap_query.as_mut() {
+            interval.set(tokio::time::interval(duration));
         }
+
+        Ok(())
     }
 }
 
--- a/crates/topos-p2p/src/behaviour/discovery.rs
+++ b/crates/topos-p2p/src/behaviour/discovery.rs
@@ -82,7 +82,7 @@ impl DiscoveryBehaviour {
             health_status: if known_peers.is_empty() {
                 HealthStatus::Healthy
             } else {
-                HealthStatus::default()
+                HealthStatus::Initializing
             },
         }
     }
--- a/crates/topos-p2p/src/behaviour/discovery.rs
+++ b/crates/topos-p2p/src/behaviour/discovery.rs
@@ -74,8 +74,16 @@ impl DiscoveryBehaviour {
         Self {
             inner: kademlia,
             current_bootstrap_query_id: None,
-            next_bootstrap_query: Some(Box::pin(tokio::time::interval(config.bootstrap_interval))),
-            health_status: Default::default(),
+            next_bootstrap_query: if known_peers.is_empty() {
+                None
+            } else {
+                Some(Box::pin(tokio::time::interval(config.bootstrap_interval)))
+            },
+            health_status: if known_peers.is_empty() {
+                HealthStatus::Healthy
+            } else {
+                HealthStatus::default()
+            },
         }
     }
 
--- a/crates/topos-p2p/src/behaviour/gossip.rs
+++ b/crates/topos-p2p/src/behaviour/gossip.rs
@@ -1,23 +1,29 @@
 use std::collections::hash_map::DefaultHasher;
+use std::collections::HashSet;
 use std::hash::{Hash, Hasher};
 use std::{
-    collections::{HashMap, HashSet, VecDeque},
+    collections::{HashMap, VecDeque},
     env,
     task::Poll,
     time::Duration,
 };
 
+use libp2p::swarm::{ConnectionClosed, FromSwarm};
+use libp2p::PeerId;
 use libp2p::{
-    gossipsub::{self, IdentTopic, Message, MessageAuthenticity, MessageId},
+    gossipsub::{self, IdentTopic, Message, MessageAuthenticity},
     identity::Keypair,
     swarm::{NetworkBehaviour, THandlerInEvent, ToSwarm},
 };
 use prost::Message as ProstMessage;
 use topos_core::api::grpc::tce::v1::Batch;
-use topos_metrics::{P2P_DUPLICATE_MESSAGE_ID_RECEIVED_TOTAL, P2P_GOSSIP_BATCH_SIZE};
-use tracing::{debug, error, warn};
+use topos_metrics::P2P_GOSSIP_BATCH_SIZE;
+use tracing::{debug, error};
 
+use crate::error::P2PError;
 use crate::{constants, event::ComposedEvent, TOPOS_ECHO, TOPOS_GOSSIP, TOPOS_READY};
+
+use super::HealthStatus;
 
 const MAX_BATCH_SIZE: usize = 10;
 
@@ -26,7 +32,10 @@ pub struct Behaviour {
     gossipsub: gossipsub::Behaviour,
     pending: HashMap<&'static str, VecDeque<Vec<u8>>>,
     tick: tokio::time::Interval,
-    cache: HashSet<MessageId>,
+    /// List of connected peers per topics
+    connected_peer: HashMap<&'static str, HashSet<PeerId>>,
+    /// The health status of the gossip behaviour
+    pub(crate) health_status: HealthStatus,
 }
 
 impl Behaviour {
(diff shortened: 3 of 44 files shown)


crates/topos-p2p/src/runtime/handle_event.rs
```diff
@@ -67,19 +71,64 @@ impl
...
+                .successfully_connect_to_bootpeer
+                .is_none()
+                && self.state_machine.dialed_bootpeer.contains(&connection_id) =>
+            {
+                warn!("Unable to connect to bootpeer {peer_id}: {error:?}");
+                self.state_machine.dialed_bootpeer.remove(&connection_id);
+                if self.state_machine.dialed_bootpeer.is_empty() {
+                    // We tried to connect to all bootnode without success
+                    error!("Unable to connect to any bootnode");
+                }
```
dvdplm, line 90: Shouldn't we retry (with exponential backoff) here?

dvdplm: To expand on that, the goal is to make it so that the TCE can get up and running regardless of the order in which nodes start/stop, and to recover from various adverse network conditions. So what if the bootnodes are busy restarting?

crates/topos-p2p/src/runtime/handle_event.rs
```diff
@@ -67,19 +71,64 @@ impl
...
 
+            SwarmEvent::ConnectionEstablished {
+                peer_id,
+                connection_id,
+                endpoint,
+                num_established,
+                concurrent_dial_errors,
+                established_in,
+            } if self.state_machine.dialed_bootpeer.contains(&connection_id) => {
+                info!("Successfully connect to bootpeer {peer_id}");
```
dvdplm, line 107:
```suggestion
                info!("Successfully connected to bootpeer {peer_id}");
```

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -28,60 +40,99 @@ pub struct Runtime {
...
 
 mod handle_command;
 mod handle_event;
 
+/// Internal state machine of the p2p layer
+///
+/// This struct may change in the future to be more flexible and to handle more
+/// complex state transitions/representation.
+#[derive(Default)]
+pub(crate) struct StateMachine {
```
dvdplm, line 58: This isn't really a state machine. It's more of a metadata container that holds information on the state of the network. I'd rename it to `NetworkState` or `ConnectivityProperties` or just `State`.

crates/topos-p2p/src/runtime/handle_event.rs
```diff
@@ -67,19 +71,64 @@ impl
...
+                if self
+                    .state_machine
+                    .successfully_connect_to_bootpeer
+                    .is_none()
+                {
+                    self.state_machine.successfully_connect_to_bootpeer = Some(connection_id);
+                    _ = self.state_machine.dialed_bootpeer.remove(&connection_id);
```
dvdplm, line 114: Is there any risk of a race condition here? I think not but better be sure...

crates/topos-p2p/src/runtime/handle_event/gossipsub.rs
```diff
@@ -9,12 +9,12 @@ use crate::{constants, event::GossipEvent, Event, Runtime, TOPOS_ECHO, TOPOS_GOS
...
 
-use super::EventHandler;
+use super::{EventHandler, EventResult};
 
 #[async_trait::async_trait]
 impl EventHandler<GossipEvent> for Runtime {
-    async fn handle(&mut self, event: GossipEvent) {
-        if let GossipEvent {
+    async fn handle(&mut self, event: GossipEvent) -> EventResult {
+        if let GossipEvent::Message {
```
dvdplm, line 17: Do we expect to have more than one variant of `GossipEvent` down the line?

crates/topos-p2p/src/config.rs
```diff
@@ -32,6 +32,9 @@ pub struct DiscoveryConfig {
     pub provider_publication_interval: Option<Duration>,
     /// Interval at which the node will send bootstrap query to the network
     pub bootstrap_interval: Duration,
+    /// Interval at which the node will send fast bootstrap query to the network
+    /// Mostly used when the node is bootstrapping and failed to connect to boot peers
```
dvdplm, line 36:
```suggestion
    /// Mostly used when the node is bootstrapping and failed to connect to boot peers.
	/// Defaults to 5 sec (['FAST_BOOTSTRAP_INTERVAL`]).
```

(Although I'm not sure if that constant should be `pub`)

crates/topos-p2p/src/config.rs
```diff
@@ -32,6 +32,9 @@ pub struct DiscoveryConfig {
     pub provider_publication_interval: Option<Duration>,
     /// Interval at which the node will send bootstrap query to the network
```
dvdplm, line 33: State default here pls

crates/topos-p2p/src/runtime/handle_event/discovery.rs
```diff
@@ -35,74 +32,106 @@ impl EventHandler<Box<Event>> for Runtime {
...
+                step,
+            } if num_remaining == 0
+                && self
+                    .state_machine
+                    .successfully_connect_to_bootpeer
+                    .is_none()
+                && self.swarm.behaviour().discovery.health_status == HealthStatus::Unhealthy =>
+            {
+                warn!(
+                    "Bootstrap query finished but unable to connect to bootnode {:?} {:?}",
```
dvdplm, line 76: Same log as above, that could be annoying. Can you change it to reflect the two cases accurately?

crates/topos-p2p/src/runtime/handle_event/discovery.rs
```diff
@@ -35,74 +32,106 @@ impl EventHandler<Box<Event>> for Runtime {
...
+                stats,
+                step,
+            } if num_remaining == 0
+                && self
+                    .state_machine
+                    .successfully_connect_to_bootpeer
+                    .is_some()
+                && self.swarm.behaviour().discovery.health_status == HealthStatus::Unhealthy =>
+            {
+                warn!(
```
dvdplm, line 110: Why is this a warning?

crates/topos-tce-api/src/runtime/sync_task.rs
```diff
@@ -80,7 +80,7 @@ impl SyncTask {
             status: SyncTaskStatus::Running,
             stream_id,
             target_subnet_stream_positions,
-            storage,
+            store: storage,
```
dvdplm, line 83: Why not rename the argument to `store` as well?

crates/topos-p2p/src/behaviour/discovery.rs
```diff
@@ -69,7 +74,16 @@ impl DiscoveryBehaviour {
         Self {
             inner: kademlia,
             current_bootstrap_query_id: None,
-            next_bootstrap_query: Some(Box::pin(tokio::time::interval(config.bootstrap_interval))),
+            next_bootstrap_query: if known_peers.is_empty() {
```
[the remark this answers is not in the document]
Freyskeyd: The `know_peers` could be either bootnodes from `genesis` file or defined in config.

The `next_bootstrap_query` is the inteval trigger of the `bootstrap` query used in the `poll` function. For `genesis` bootnode I encounter the fact that if they're alone at start, they can't send any `bootstrap` query because they know no one. As a quick win I just disable the `bootstrap` queries for lonely bootnode, meaning that they will not try to walk the network but still receive connection and aggregate those connection.

This is only possible for `bootnode` as the `known_peers` have the list of bootnode from the genesis file, during the configuration parse, the local peer_id is excluded from this list. If the local node is the only bootnode in the genesis file, the list is then empty but for other node nothing change.

---

> Both of next_bootstrap_query and health_status are confusing to me. No known peers means we're healthy?

`next_bootstrap_query` is `The next bootstrap query interval used to schedule the next bootstrap query` as define above

`health_status` is the current heath status of the behaviour based on the event it consumes.

For a lonely bootnode that doesn't need any discovery the `HealthStatus` is `healthy` by default. For any other node the healthy status is reached when the first `bootstrap` query to resolve with one of the defined `bootnode` are successful.

crates/topos-p2p/src/config.rs
Freyskeyd: What do you mean?

crates/topos-p2p/src/config.rs
Freyskeyd: This `DiscoverConfig` may end up in `topos-config` so I'm ok to let it public

crates/topos-p2p/src/runtime/handle_event.rs
Freyskeyd: This PR is to allow nodes to start in any order. The recovering or any other behaviour are not handled yet.

The retries are handled not manually but rely on the frequency of the `bootstrap` queries.
The dialling of the bootnode is triggered by the initialization of a bootstrap query.
The `dialed_bootpeer` is handling single dialing for one bootnode as represented by the `ConnectionId`.

crates/topos-p2p/src/runtime/handle_event/gossipsub.rs
Freyskeyd: Yes we do, mostly to deal with errors

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -20,6 +29,9 @@ pub struct Runtime {
     pub(crate) listening_on: Vec<Multiaddr>,
     pub(crate) public_addresses: Vec<Multiaddr>,
 
+    /// Boot peers to connect used to bootstrap the p2p layer
```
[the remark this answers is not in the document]
Freyskeyd: That's not just well-know public bootnode, that's any bootnode. It could be a single private `sentry` node for a protected validator.

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -28,60 +40,99 @@ pub struct Runtime {
...
+    /// Internal state machine of the p2p layer
+    pub(crate) state_machine: StateMachine,
+
+    pub(crate) health_status: HealthStatus,
 }
 
 mod handle_command;
 mod handle_event;
 
+/// Internal state machine of the p2p layer
```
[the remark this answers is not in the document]
Freyskeyd: It's internal to the `crate`, `pub(crate)` make the struct visible within the current crate.

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -28,60 +40,99 @@ pub struct Runtime {
...
+#[derive(Default)]
+pub(crate) struct StateMachine {
+    /// Indicates if the node has external addresses configured
+    pub(crate) has_external_addresses: bool,
+    /// Indicates if the node is listening on any address
+    pub(crate) is_listening: bool,
+    /// List the boot peers that the node has tried to connect to
+    pub(crate) dialed_bootpeer: HashSet<ConnectionId>,
+    /// Indicates if the node has successfully connected to a boot peer
+    pub(crate) successfully_connect_to_bootpeer: Option<ConnectionId>,
```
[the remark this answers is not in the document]
Freyskeyd: It is only set once. We just care of knowing if we were able to connect to at least one bootnode

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -28,60 +40,99 @@ pub struct Runtime {
...
+    /// Track the number of retries to connect to boot peers
```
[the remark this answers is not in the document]
Freyskeyd: It's the number of retries the node has to connect to any boot peers. If it has 5 retries and failed to connect 3 times to any bootnode it has 2 more attempt for the next bootstrap queries.

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -28,60 +40,99 @@ pub struct Runtime {
...
+                result = &mut handle => {
+                    match result {
+                        Ok(Ok(_)) => info!("P2P layer has been shutdown"),
+                        Ok(Err(error)) => {
+                            error!("P2P layer has failed with error: {:?}", error);
+
+                            return Err(error);
+                        }
+                        Err(_) => {
+                            error!("P2P layer has failed in an unexpected way.");
```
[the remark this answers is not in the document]
Freyskeyd: Because it's a `JoinError` from `tokio`, it doesn't bring much information you will endup with something like:

`P2P layer has failed in an unexpected way: JoinError::Panic(Id(X), ...)`

or

`P2P layer has failed in an unexpected way: "task X panicked"`

with the panic message being print before in `stderr`

crates/topos-config/src/edge/command.rs
```diff
@@ -64,6 +64,7 @@ impl CommandConfig {
     }
 
     pub async fn spawn(self) -> Result<ExitStatus, std::io::Error> {
+        info!("Spawning Polygon Edge with args: {:?}", self.binary_path);
```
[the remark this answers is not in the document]
Freyskeyd: This is specifically spawning `polygon-edge` and the command is in the `edge` module so it makes sense to keep the log that way I guess.

crates/topos-p2p/src/runtime/mod.rs
```diff
@@ -95,4 +146,29 @@ impl Runtime {
...
+        let new_status = match (discovery, gossipsub) {
+            (HealthStatus::Initializing, _) | (_, HealthStatus::Initializing) => {
+                HealthStatus::Initializing
+            }
+            (HealthStatus::Unhealthy, _) | (_, HealthStatus::Unhealthy) => HealthStatus::Unhealthy,
+            (HealthStatus::Recovering, _) | (_, HealthStatus::Recovering) => {
+                HealthStatus::Recovering
+            }
+            (HealthStatus::Healthy, HealthStatus::Healthy) => HealthStatus::Healthy,
+        };
```
hadjiszs, line 164: If we have one `initializing` and the other `unhealthy` that will be the first match so we'll consider being `initializing`? If so maybe we wanna reflect the "unhealthyness" as top priority by having it as 1st branch?

hadjiszs (commented):
Also saw several occurrences "bootpeers" and "bootnodes" maybe let's try to stick to one or you wanna actually keep distinction for some reason?

crates/topos-p2p/src/runtime/mod.rs
Freyskeyd: Yeah, on the `behaviour` I was thinking that we could wait for every behaviour to finish the `initialization` phase before deciding if the `layer` is `unhealthy`. But we may need another variant such as `Killing` defining that the node need to kill itself.

crates/topos-p2p/src/behaviour/discovery.rs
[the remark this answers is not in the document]
dvdplm: Cool, thank you for explaining, much clearer now. Can you add a comment explaining that "Bootnodes do not have any known peers, so we skip bootstrapping for them" or something similar?

crates/topos-p2p/src/runtime/mod.rs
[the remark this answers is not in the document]
dvdplm:
```suggestion
    /// Well-known (or private) boot peers used to bootstrap the p2p layer.
```

crates/topos-p2p/src/runtime/mod.rs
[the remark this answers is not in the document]
dvdplm: I'd add this to the docs then: "…has successfully connected to at least one bootnode".

dvdplm (approved):

crates/topos-p2p/src/behaviour/gossip.rs
```diff
@@ -186,9 +212,62 @@ impl NetworkBehaviour for Behaviour {
...
+                            },
+                        )))
+                    }
+                    _ => {}
+                },
+                gossipsub::Event::Subscribed { peer_id, topic } => {
+                    debug!("Subscribed to {:?} with {peer_id}", topic);
+                    if self.health_status != HealthStatus::Healthy {
+                        self.health_status = HealthStatus::Healthy;
+                    }
```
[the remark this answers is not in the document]
hadjiszs: >For simplicity first. We could add more later logic on this if we want but as one peer always connect to the three topics... What do you think?

Just had a slight concern on the case where we are considered as healthy and so starting to receive payload while we are not subscribed yet to some topics for instance. Anyway I see you changed to consider the 3 topics :ok_hand:

PR #465, opened 2024-02-27
chore: refactor logs and fix typo

hadjiszs:
# Description

Simple PR to revise the level of some logs and fix some typo.

## PR Checklist:

### DIFF
--- a/crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
+++ b/crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
@@ -12,7 +12,7 @@ use topos_core::{
 };
 use topos_crypto::messages::MessageSigner;
 use topos_metrics::DOUBLE_ECHO_BROADCAST_FINISHED_TOTAL;
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, trace};
 mod status;
 
 pub use status::Status;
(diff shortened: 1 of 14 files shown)


crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
```diff
@@ -66,7 +66,7 @@ impl BroadcastState {
         });
 
         if need_gossip {
-            warn!("📣 Gossiping the Certificate {}", &state.certificate.id);
+            debug!("📣 Gossiping the Certificate {}", &state.certificate.id);
```
dvdplm, line 69: It would be good to log the subnet involved here, and perhaps the validator id too?

dvdplm (approved):

crates/topos-tce-broadcast/src/double_echo/broadcast_state.rs
hadjiszs: Yep, added subnet-id here https://github.com/topos-protocol/topos/pull/465/commits/7e4f32eb41aed838815ebfb6704f0cc78d816c28

As for the `validator-id`, I might prefer instead to have a robust and complete solution to expose the mapping `instance <> validator-id` for all the logs instead of sparkling some `validator-id` on some logs here and there no?

PR #467, opened 2024-02-27
fix: concurrency insert between pending and delivered

Freyskeyd:
# Description

This PR is covering a corner case which involve the delivery of a prev certificate and a child certificate.

The scenario is this one:

- A `prev` certificate (`C1`) has been delivered (by the broadcast) and need to be persisted
  The persist method will hold a lock while performing multiple insert/delete to avoid
  insert race condition.
- At the same time, another node is sending a certificate (`C2`) which have `C1` as `prev_id`.
  `C2` is looking at the storage to find if the `prev_id` `C1` is delivered but find nothing as
  the `persist` method is still working at creating the `WriteBatch`. It led the node to put
  `C2` in the `precedence_pool` waiting for `C1` to be delivered while it is in fact already
  delivered.

To avoid that and as a first step, when trying to insert a certificate in the pending pool,
The node will try to acquire a lock guard on the certificate but also on the prev_id.

## PR Checklist:

### DIFF
--- a/crates/topos-tce-broadcast/src/tests/task.rs
+++ b/crates/topos-tce-broadcast/src/tests/task.rs
@@ -1,4 +1,4 @@
-use std::{future::IntoFuture, str::FromStr, sync::Arc, time::Duration};
+use std::{future::IntoFuture, sync::Arc, time::Duration};
 
 use rstest::rstest;
 use tokio::{
@@ -11,6 +11,7 @@ use topos_tce_storage::validator::ValidatorStore;
 use topos_test_sdk::{
     certificates::create_certificate_chain,
     constants::{SOURCE_SUBNET_ID_1, TARGET_SUBNET_ID_1},
+    crypto::message_signer,
     storage::create_validator_store,
 };
 
@@ -26,6 +27,7 @@ async fn start_with_ungossiped_cert(
     #[future(awt)]
     #[from(create_validator_store)]
     validatore_store: Arc<ValidatorStore>,
+    message_signer: Arc<MessageSigner>,
 ) {
     let certificate = create_certificate_chain(SOURCE_SUBNET_ID_1, &[TARGET_SUBNET_ID_1], 1)
         .pop()
@@ -40,10 +42,6 @@ async fn start_with_ungossiped_cert(
     };
     let (event_sender, mut event_receiver) = mpsc::channel(2);
     let (broadcast_sender, _) = broadcast::channel(1);
-    let message_signer = Arc::new(
-        MessageSigner::from_str("122f3ae6ade1fd136b292cea4f6243c7811160352c8821528547a1fe7c459daf")
-            .unwrap(),
-    );
     let need_gossip = true;
     let subscriptions = SubscriptionsView::default();
 
--- a/crates/topos-tce-broadcast/src/tests/task.rs
+++ b/crates/topos-tce-broadcast/src/tests/task.rs
@@ -1,0 +1,86 @@
+use std::{future::IntoFuture, str::FromStr, sync::Arc, time::Duration};
+
+use rstest::rstest;
+use tokio::{
+    spawn,
+    sync::{broadcast, mpsc},
+};
+use topos_core::uci::Certificate;
+use topos_crypto::{messages::MessageSigner, validator_id::ValidatorId};
+use topos_tce_storage::validator::ValidatorStore;
+use topos_test_sdk::{
+    certificates::create_certificate_chain,
+    constants::{SOURCE_SUBNET_ID_1, TARGET_SUBNET_ID_1},
+    storage::create_validator_store,
+};
+
+use crate::{
+    double_echo::broadcast_state::BroadcastState, event::ProtocolEvents,
+    sampler::SubscriptionsView, task_manager::task::Task,
+};
+
+#[rstest]
+#[test_log::test(tokio::test)]
+#[timeout(Duration::from_secs(1))]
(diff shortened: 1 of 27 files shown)


crates/topos-tce-storage/src/tests/pending_certificates.rs
```diff
@@ -84,5 +87,128 @@ async fn adding_pending_certificate_already_delivered(store: Arc<ValidatorStore>
 
     assert!(store
         .insert_pending_certificate(&initial_certificate_delivered.certificate)
+        .await
         .is_err());
 }
+
+/// This test is covering a corner case which involve the delivery of a prev certificate
```
dvdplm, line 94:
```suggestion
/// This test is covering a corner case which involves the delivery of a prev certificate
```

crates/topos-tce-storage/src/validator/mod.rs
```diff
@@ -286,10 +286,18 @@ impl ValidatorStore {
         Ok(ids)
     }
 
-    pub fn insert_pending_certificate(
+    pub async fn insert_pending_certificate(
         &self,
         certificate: &Certificate,
     ) -> Result<Option<PendingCertificateId>, StorageError> {
+        // A lock guard is asked during the insertion of a pending certificate
```
dvdplm, line 293:
```suggestion
        // A lock guard is taken during the insertion of a pending certificate
```

crates/topos-tce-storage/src/validator/mod.rs
```diff
@@ -286,10 +286,18 @@ impl ValidatorStore {
...
+        // to avoid race condition when a certificate is being inserted and added
+        // to the pending pool at the same time
```
dvdplm, line 295:
```suggestion
        // to avoid race condition when a certificate (C1) is delivered and is in the process of being persisted to disk (so no longer pending), and a dependent certificate (C2) is added
        // to the pending pool even though it could be delivered as well.
```

crates/topos-tce-storage/src/validator/mod.rs
```diff
@@ -312,6 +320,14 @@ impl ValidatorStore {
             ));
         }
 
+        // A lock guard is asked during the insertion of a pending certificate
```
dvdplm, line 323:
```suggestion
        // A lock guard is taken during the insertion of a pending certificate
```

crates/topos-tce-broadcast/src/tests/task_manager.rs
```diff
@@ -19,8 +19,11 @@ use crate::{sampler::SubscriptionsView, task_manager::TaskManager};
 
 #[rstest]
 #[tokio::test]
-async fn can_start(#[future] create_validator_store: Arc<ValidatorStore>) {
-    let validator_store = create_validator_store.await;
+async fn can_start(
+    #[future(awt)]
```
dvdplm, line 23: What does this do?

dvdplm (approved):
LGTM, and thank you for the high quality comments explaining what is going on.

Beyond this PR, I think this need for strong synchronization is unfortunate. It makes the code less clear and introduces delays that could harm performance. Not sure how we could do better but I'll try to give it some thought. I have a gut feeling that treating pending and delivered certificates separately is part of the problem, but I'll have to think more on this.

gruberb (approved):

crates/topos-tce-broadcast/src/tests/task.rs
```diff
@@ -0,0 +1,86 @@
...
+    let validator_id = ValidatorId::default();
+    let thresholds = topos_config::tce::broadcast::ReliableBroadcastParams {
+        echo_threshold: 1,
+        ready_threshold: 1,
+        delivery_threshold: 1,
+    };
+    let (event_sender, mut event_receiver) = mpsc::channel(2);
+    let (broadcast_sender, _) = broadcast::channel(1);
+    let message_signer = Arc::new(
+        MessageSigner::from_str("122f3ae6ade1fd136b292cea4f6243c7811160352c8821528547a1fe7c459daf")
```
atanmarko, line 44: Could you put it in `topos_test_sdk::constants`?

atanmarko (approved):

crates/topos-tce-storage/src/validator/mod.rs
Freyskeyd: That's not quit true but I will rephrase

crates/topos-tce-broadcast/src/tests/task_manager.rs
Freyskeyd: it awaits the future for us

crates/topos-tce-storage/src/fullnode/mod.rs
```diff
@@ -82,6 +83,27 @@ impl FullNodeStore {
             index_tables,
         }))
     }
+
+    /// Await for a [`LockGuards`] for the given certificate id
+    pub(crate) async fn get_certificate_lock_guard(
```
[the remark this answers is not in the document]
Freyskeyd: For me we want to get a `lock_guard` for a particular certificate id from what's close to a `HashMap`, what are your arguments not to use `get_` here?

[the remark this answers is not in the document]
dvdplm: "get" is implied, you don't really need it. And it's not idiomatic Rust (see the [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#c-getter)).

PR #471, opened 2024-03-01
fix: block handling during certificate generation

atanmarko:
# Description

Handle removal of subnet blocks from the list when generating certificates

Fixes TP-786

## PR Checklist:


crates/topos-sequencer-subnet-runtime/src/certification.rs
```diff
@@ -52,8 +53,11 @@ impl Certification {
         let subnet_id = self.subnet_id;
         let mut generated_certificates = Vec::new();
 
+        // Keep account of blocks with generated certificates so that we can remove them from
+        // finalized blocks
+        let mut processed_blocks: Vec<u64> = Vec::with_capacity(self.finalized_blocks.len());
```
Freyskeyd, line 58: This allocation will be very big at some point right?

crates/topos-sequencer-subnet-runtime/src/certification.rs
```diff
@@ -126,7 +131,27 @@ impl Certification {
         }
 
         // Remove processed blocks
-        self.finalized_blocks.clear();
+        for processed_block_number in processed_blocks {
```
Freyskeyd, line 134: I don't understand this code block. When I read the code above I see that every `block_number` are added to `processed_blocks`, if an error happens then this code block isn't even executed. Am I missing something?

crates/topos-sequencer-subnet-runtime/src/certification.rs
atanmarko: `finilized_blocks` are only blocks for which certificates are not generated. There could be maybe up to tens of thousands of blocks if TCE network is not available for days so not a big deal

crates/topos-sequencer-subnet-runtime/src/certification.rs
atanmarko: Yes, we only want to remove block from `self.finalized_blocks` if the certificate is generated for that block. Otherwise, it will be processed again on the next certificate call.  `processed_blocks` keeps the blocks for which the certificate is generated

crates/topos-sequencer-subnet-runtime/src/lib.rs
```diff
@@ -18,6 +18,8 @@ pub mod proxy;
 
 use crate::proxy::{SubnetRuntimeProxyCommand, SubnetRuntimeProxyEvent};
 
+const EVENT_SUBSCRIBER_CHANNEL_SIZE: usize = 64;
```
dvdplm, line 21: We should have a comment here about the reasoning for chosing `64` and not some other number; even if it's just "off the top of my head".

crates/topos-sequencer-subnet-runtime/src/certification.rs
```diff
@@ -19,6 +18,8 @@ pub struct Certification {
     signing_key: Vec<u8>,
     /// Optional synchronization from particular block number
     pub start_block: Option<u64>,
+    /// Latest BLOCK_HISTORY_LENGTH blocks kept in memory
```
dvdplm, line 21:
```suggestion
    /// We keep [`BLOCK_HISTORY_LENGTH`] blocks in memory to check that the we process the block in the right order.
```

…or something that explains what this actually does.

crates/topos-sequencer-subnet-runtime/src/certification.rs
dvdplm: Isn't finalized_blocks hardcoded to 256? (`BLOCK_HISTORY_LENGTH`)

crates/topos-sequencer-subnet-runtime/src/certification.rs
dvdplm: Wouldn't `certified_blocks` be a better name than `processed_blocks` then?

crates/topos-tce-proxy/src/client.rs
```diff
@@ -381,6 +381,7 @@ impl TceClientBuilder {
                                 let tce_endpoint = tce_endpoint.clone();
                                 let tce_grpc_client = tce_grpc_client.clone();
                                 let context_backoff = context.clone();
+                                // TODO: Push certificates to the TCE one by one
```
dvdplm, line 384: I don't understand this TODO. I think it would be better to create a proper ticket with a full description.

crates/topos-sequencer-subnet-runtime/src/certification.rs
atanmarko: @dvdplm It was, and then we start dropping blocks if there are more than 256 pending. So this was a bug.

crates/topos-sequencer-subnet-runtime/src/certification.rs
atanmarko: Updated comment, this one was outdated

dvdplm (approved):

dvdplm:
I don't fully understand why we keep the finalized blocks in linked list as opposed to a vec or a hashmap, but that's not the main point of this pr.

atanmarko:
> I don't fully understand why we keep the finalized blocks in linked list as opposed to a vec or a hashmap, but that's not the main point of this pr.

We remove blocks from the beginning and add them to the end of the list and they need to keep order. I believe a linked list is the most optimal structure for this.

PR #483, opened 2024-03-21
fix(p2p): rework ticks of bootstrap query interval

Freyskeyd:
# Description

This PR reworks how we manage the ticks of the bootstrap queries.

After investigating an issue on `devnet` I saw that two queries were executed at the same second.
While the expected behavior is to have a delay of 60sec (by default).

### Why ?

```
09:33:29: DEBUG -> Started kademlia bootstrap query with query_id: QueryId(0)
                    ^ (1) This is the first query is created when the node starts
09:33:29: INFO  -> Dialing bootpeer 16Uiu2HAmTqPREKjx7dYpcEsjByzy8yZMgjipzAcStMFQBAvZZL73 on connection: 1
                    ^ We dial the other peer for this query
09:33:29: WARN  -> Unable to connect to bootpeer 16Uiu2HAmTqPREKjx7dYpcEsjByzy8yZMgjipzAcStMFQBAvZZL73
                    ^ The other bootnode is not available
09:33:29: WARN  -> Bootstrap query finished but unable to connect to bootnode during initialization, switching to unhealthy and fast bootstrap mode
                                    ^ The query finished but the bootnode was unreachable, switching to fast mode (5sec interval query)
09:33:29: INFO  -> Dialing bootpeer 16Uiu2HAmTqPREKjx7dYpcEsjByzy8yZMgjipzAcStMFQBAvZZL73 on connection: 2
                    ^ (2) New query created !
09:33:29: WARN  -> Unable to connect to bootpeer 16Uiu2HAmTqPREKjx7dYpcEsjByzy8yZMgjipzAcStMFQBAvZZL73
09:33:29: WARN  -> Bootstrap query finished but unable to connect to bootnode, retrying 2 more times
```

As we can see on the logs, the first query `(1)` fails, and we switch to fast mode for bootstrap as expected.
But the problem is that after changing the mode (basically, creating a new `interval` with shorter period), the next query `(2)` is directly executed.

This is due to the fact that the first `tick` of `interval` completes immediately and then the discovery `poll` function fire a new query.

### Fix

To fix that I updated the `change_interval` to await and ignore the first tick and switch the `MissedTickBehavior` to `delay` as we don't need any catchup on missing tick.

## PR Checklist:

### DIFF
--- a/crates/topos-p2p/src/behaviour/discovery.rs
+++ b/crates/topos-p2p/src/behaviour/discovery.rs
@@ -80,7 +80,10 @@ impl DiscoveryBehaviour {
             next_bootstrap_query: if known_peers.is_empty() {
                 None
             } else {
-                Some(Box::pin(tokio::time::interval(config.bootstrap_interval)))
+                let mut interval = tokio::time::interval(config.bootstrap_interval);
+                interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+
+                Some(Box::pin(interval))
             },
             health_status: if known_peers.is_empty() {
                 HealthStatus::Healthy
@@ -105,9 +108,14 @@ impl DiscoveryBehaviour {
     }
 
     /// Change the interval of the next bootstrap queries
-    pub fn change_interval(&mut self, duration: Duration) -> Result<(), P2PError> {
+    pub async fn change_interval(&mut self, duration: Duration) -> Result<(), P2PError> {
         if let Some(interval) = self.next_bootstrap_query.as_mut() {
-            interval.set(tokio::time::interval(duration));
+            let mut new_interval = tokio::time::interval(duration);
+            // Delay the next tick
+            new_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+            // ignore first tick
+            _ = new_interval.tick().await;
+            interval.set(new_interval);
         }
 
         Ok(())
--- a/crates/topos-p2p/src/behaviour.rs
+++ b/crates/topos-p2p/src/behaviour.rs
@@ -9,7 +9,7 @@ pub(crate) mod peer_info;
 pub(crate) mod topos;
 
 /// Represents the health status of a behaviour inside the p2p layer
-#[derive(Default, PartialEq, Eq)]
+#[derive(Debug, Default, PartialEq, Eq)]
 pub(crate) enum HealthStatus {
     #[default]
     Initializing,
--- a/crates/topos-p2p/src/runtime/handle_event.rs
+++ b/crates/topos-p2p/src/runtime/handle_event.rs
@@ -197,6 +197,7 @@ impl EventHandler<SwarmEvent<ComposedEvent>> for Runtime {
             _ = self.event_sender.send(event).await;
         }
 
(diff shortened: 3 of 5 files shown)


crates/topos-p2p/src/behaviour/discovery.rs
```diff
@@ -80,7 +80,10 @@ impl DiscoveryBehaviour {
             next_bootstrap_query: if known_peers.is_empty() {
                 None
             } else {
-                Some(Box::pin(tokio::time::interval(config.bootstrap_interval)))
+                let mut interval = tokio::time::interval(config.bootstrap_interval);
+                interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
```
dvdplm, line 84: So the default is `Burst` right? (ref for others: https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html#variants)

crates/topos-p2p/src/behaviour/discovery.rs
```diff
@@ -105,9 +108,14 @@ impl DiscoveryBehaviour {
...
     /// Change the interval of the next bootstrap queries
-    pub fn change_interval(&mut self, duration: Duration) -> Result<(), P2PError> {
+    pub async fn change_interval(&mut self, duration: Duration) -> Result<(), P2PError> {
         if let Some(interval) = self.next_bootstrap_query.as_mut() {
-            interval.set(tokio::time::interval(duration));
+            let mut new_interval = tokio::time::interval(duration);
+            // Delay the next tick
+            new_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+            // ignore first tick
+            _ = new_interval.tick().await;
```
dvdplm, line 117: In theory this isn't really needed right? We could also just let it tick and fire off the first query right away without pausing? (Not suggesting we do, but just to understand)

dvdplm (approved):
LGTM.

> retrying 2 more times

One question on this: why just 3 retries? Ideally we'd keep retrying (perhaps backing off over time).

crates/topos-p2p/src/behaviour/discovery.rs
Freyskeyd: If our first query is completing (failing) at `10s`, we change the interval and we expect the `Q2` to execute `5sec` later. Not directly as we want to let time to the network to change a bit (node going up in the delay).

```
                  0        10s   15s   20s   25s   30s
Expected Queries: |   Q1     | --- | Q2  | Q3  | Q4  | Q5  |
Actual Queries:   |   Q1     | Q2  | Q3  | Q4  | Q5  |
Mode:             | Normal   |  Fast
```

crates/topos-p2p/src/behaviour/discovery.rs
Freyskeyd: yes it is

hadjiszs (approved):

crates/topos-p2p/src/behaviour/discovery.rs
dvdplm: Yes, I think I understand that part. But if the number of retries is decently large, or even infinite, then letting the first one fire right away wouldn't be a problem right?

